PR #4, opened 2016-11-18
Support all supported ruby versions

teoljungberg:
Which currently is 2.1.*, 2.2.*, 2.3.*

- Test against all supported rubies on Travis.

  - Allow 2.1.* to fail against Rails 5.0 since Rails requires 2.2.* or
  above.

- Remove usage of squiggly-heredocs(`<<~`).

  Since we want to support all supported version of Ruby, we have to use
  features that are available between all of them. Squiggly-heredocs are
  only available in 2.3.* and above.

- Simplify tests for:

  - `Fx::Adapters::Postgres::Functions#functions`
  - `Fx::Adapters::Postgres::Triggers#triggers`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,8 @@ notifications:
   email: false
 rvm:
   - 2.3.1
+  - 2.2.6
+  - 2.1.10
 gemfile:
   - gemfiles/rails40.gemfile
   - gemfiles/rails41.gemfile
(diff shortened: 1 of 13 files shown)


arunagw:
LGTM

tappleby:
👍  Good timing, I forked this library last night just to try it out on ruby 2.2!

image

teoljungberg:
@tappleby It was your fork that made me go through this today. So thank you!

teoljungberg:
I also released v0.3.0 with support for 2.1 and 2.2 @tappleby

Issue #7, opened 2017-01-17
Can't use function with parameters


timothee-alby:
Thanks for this, as  a user of [scenic](https://github.com/thoughtbot/scenic), I was happy to find a similar tool for functions :)

Unfortunately, this doesn't work for functions with parameters: although creating function isn't a problem, dropping function isn't possible so migrations can't be reverted.

I believe the problem comes from [Fx::Adapters::Postgres:: drop_function](https://github.com/teoljungberg/fx/blob/master/lib/fx/adapters/postgres.rb#L127-L129) as the brakets (`()`) are added automatically. 
I have functions with parameters and the function signature must be specified to drop it (because one can create different functions with the same name but different parameters).

A fix could be that when the function name is specified as a string, the brackets are not added? If backward-compatibility is a concern, the alternatively could be to pass a list of parameters as an option?

teoljungberg:
Hi @timothee-alby!

That's a good use-case that I didn't consider, I first built this to resolve my issues around using triggers and functions combined. But F(x) should work with all kinds of functions.

After doing some quick reading, it seems we can leverage [pg_get_function_identity_arguments] to get the arguments and therefore we can determine the function signature as a whole. It will require a lookup in the database, but I'd deem it's necessary. What do you think?

I'll see if I can open a PR for you to try out and review to resolve this issue.

[pg_get_function_identity_arguments]: https://www.postgresql.org/docs/current/static/functions-info.html#FUNCTIONS-INFO-CATALOG-TABLE

teoljungberg:
I've opened https://github.com/teoljungberg/fx/pull/8 to hopefully resolve the issue.

timothee-alby:
Thank you, that was quick!

The problem is that I have multiple functions with the same name but different signatures (yes, this is a bit insane, I know...)
That's why I suggested passing the arguments manually

teoljungberg:
> Thank you, that was quick!

I aim to please.

> The problem is that I have multiple functions with the same name but different signatures (yes, this is a bit insane, I know...)
> That's why I suggested passing the arguments manually

That should be taken care of automatically, since the entire function signature is what we're using to identify it.
But, if you can have some varying examples I can use to sanity test this. That would be much obliged.

timothee-alby:
Say you create 2 functions with the same name:
```sql
CREATE FUNCTION foo(bar INTEGER) RETURNS INTEGER AS 'select 1' LANGUAGE SQL;
[...]

PR #10, opened 2017-07-25
Support functions with multiple definitions

jlogsdon:
This should resolve https://github.com/teoljungberg/fx/issues/7


lib/fx/adapters/postgres.rb
```diff
@@ -125,7 +125,11 @@ def update_trigger(name, on:, sql_definition:)
       #
       # @return [void]
       def drop_function(name)
-        execute "DROP FUNCTION #{name}();"
+        fns = functions.select { |function| function.name == name.to_s }
+
+        fns.each do |function|
+          execute "DROP FUNCTION #{name}(#{function.arguments});"
```
teoljungberg: What do you think introducing a `Function#signature` method? That way we could return `test()` or `test(int a)` from that method, and clarify this sql exectution?

lib/fx/adapters/postgres.rb
```diff
@@ -125,7 +125,11 @@ def update_trigger(name, on:, sql_definition:)
       #
       # @return [void]
       def drop_function(name)
-        execute "DROP FUNCTION #{name}();"
+        fns = functions.select { |function| function.name == name.to_s }
```
teoljungberg: I generally try to shy away from using acronyms as variable names. Keeping a local variable for a single loop seems redundant. So what do you think of extracting this into a private method along the lines of `functions_with_matching_signature(name)` and use that directly in the `.each` loop?

lib/fx/adapters/postgres/functions.rb
```diff
@@ -11,13 +11,16 @@ class Functions
...
               pp.proname AS name,
+              pg_get_function_identity_arguments(pp.oid) AS arguments,
               pg_get_functiondef(pp.oid) AS definition
           FROM pg_proc pp
           JOIN pg_namespace pn
               ON pn.oid = pp.pronamespace
           LEFT JOIN pg_depend pd
               ON pd.objid = pp.oid AND pd.deptype = 'e'
-          WHERE pn.nspname = 'public' AND pd.objid IS NULL
+          WHERE pn.nspname = 'public'
```
teoljungberg: Since this is covered by #11 - what do you think of excluding this in this PR as it is unrelated to this feature?

lib/fx/adapters/postgres/functions.rb
```diff
@@ -32,7 +35,8 @@ def initialize(connection)
           @connection = connection
         end
 
-        # All of the functions that this connection has defined.
+        # All of the functions that this connection has defined. Functions with
```
teoljungberg: I don't understand the added comment. Can you clarify?

lib/fx/function.rb
```diff
@@ -3,23 +3,26 @@ module Fx
...
 
     def ==(other)
-      name == other.name && definition == other.definition
+      name == other.name &&
+        definition == other.definition &&
+        arguments == other.arguments
     end
 
     def to_schema
-      <<-SCHEMA.indent(2)
```
teoljungberg: I had so many pains getting this to render nicely in the schema. Can you validate that the output is unchanged?

spec/acceptance/multiple_definition_functions_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+      RETURNS text AS $$
+      BEGIN
+          RETURN test('test default');
+      END;
+      $$ LANGUAGE plpgsql;
+    EOS
+    connection.create_function :test, sql_definition: sql_definition
+  end
+
+  it "dumps the function's to the schema" do
```
teoljungberg: `s/function's/functions`.

spec/acceptance/multiple_definition_functions_spec.rb
```diff
@@ -0,0 +1,45 @@
+require "acceptance_helper"
+
+describe "Multiple definition functions" do
```
teoljungberg: What do you think of merging this test into `spec/acceptance/user_manages_functions_spec.rb` instead?

spec/acceptance/multiple_definition_functions_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+  before(:each) do
```
teoljungberg: I generally shy away from `before`, `around` and `after` blocks. And would probably prefer if were to do so here too and inline the creation of each function in their own test.

We would decrease the duplication if the functions did not re-use themselves between each other.

spec/features/functions/multiple_identities_spec.rb
```diff
@@ -0,0 +1,42 @@
+describe "Functions with multiple definitions", :db do
```
teoljungberg: Are the identities the factor that varies? I would see it as the arguments.

Could we clarify this test description somehow?

spec/features/functions/multiple_identities_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+      END;
+      $$ LANGUAGE plpgsql;
+    EOS
+
+    with_function_definition(name: :test, sql_definition: sql_definition) do
+      example.run
+    end
+  end
+
+  let(:functions) { Fx.database.functions }
```
teoljungberg: I would inline this `let` into each test.

spec/features/functions/multiple_identities_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+    connection.create_function(:test)
+
+    expect(functions[0].name).to eql("test")
+    expect(functions[0].arguments).to eql("")
+
+    expect(functions[1].name).to eql("test")
+    expect(functions[1].arguments).to eql("str text")
+  end
+
+  it "drops both functions" do
```
teoljungberg: I do not understand the behavior in this test. Can you clarify its use somehow?

spec/features/functions/multiple_identities_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+      example.run
+    end
+  end
+
+  let(:functions) { Fx.database.functions }
+
+  it "creates both functions" do
+    connection.create_function(:test)
+
+    expect(functions[0].name).to eql("test")
```
teoljungberg: I've consistently used `.to eq("test")` rather than `.to eql("test")` as the object equallity is less important here.

teoljungberg: That we have to use `functions[0]` and `functions[1]` here is a smell to me that the state of the `functions` is varying. That is why I prefer moving away from `before` and `after` blocks, as it relies on state set elsewhere.

lib/fx/adapters/postgres.rb
jlogsdon: Great suggestion. I think my original implementation had a `functions_by_name` further up the tree that I didn't like, but a helper method locally here makes a ton of sense.

lib/fx/adapters/postgres/functions.rb
jlogsdon: Yup. This got pulled in when testing locally and I forgot to remove it when pushing.

lib/fx/adapters/postgres/functions.rb
jlogsdon: This comment is currently a lie, anyway: functions are _not_ grouped. The initial plan was for Function objects to contain all the different signatures, but that was making things too complex for no reason. I'll remove this comment, I don't believe updating it adds anything as the original comment still completely describes its behavior.

lib/fx/function.rb
jlogsdon: Yup! I actually noticed indentation issues when using this in our Rails project.

spec/acceptance/multiple_definition_functions_spec.rb
jlogsdon: Makes sense to me.

lib/fx/adapters/postgres.rb
jlogsdon: Great idea, and that gives us "signature" as a reference to this feature instead of waffling between "definition" and "arguments".

jlogsdon:
Excellent feedback! I hope to have changes ready for another review in a day or two.

jlogsdon:
I goofed by rebasing to remove the query change in [this comment](https://github.com/teoljungberg/fx/pull/10#discussion_r130222455). The rest of the PR feedback is new commits starting from 29bb372.

jlogsdon:
I'm on vacation until August 14th and won't be able to respond to any feedback on this or my other issues until then.

teoljungberg:
I'm also on vacation, but I should have gone through your issues and PRs while you are away @jlogsdon.

Thanks for all the great work!

stoivo:
ping

teoljungberg:
I'm interested in getting this merged and revised, what do you think @jlogsdon?

teoljungberg:
Hi @jlogsdon. If you don't want to continue working on this, I might take over and using this as a stepping point. What do you think?

My largest concern right now is that we are removing all functions that share the same name, when we should differentiate based on the argument signature. And if there are duplicates once we try to drop a function, we should raise an error, and allow users to provide the arguments themselves.

I'm thinking an API like:


```ruby
create_function :foo_bar, arguments: "int a"
```

md5:
@teoljungberg Have you thought any more about this? It seems like you'll need to adjust the naming convention for the `*.sql` files to include the argument types if you start supporting functions with multiple definitions.

teoljungberg:
@md5 I would love to support this, and I will add it once I personally have a need for it in the applications I work with. That said, I'd love a PR from someone else who might know more about it than myself.

teoljungberg:
Closing this PR due to inactivity.

I'll happily support this feature if a new PR is opened.

Matchlighter:
I'd very much like to see support for this as well (particularly as it fixes #7), but don't have time to submit a PR.
Here's my workaround for #7 in the meantime though:
```ruby
class CreateFunctionNaturalsort < ActiveRecord::Migration[5.0]
  def up
    create_function :naturalsort
  end

  def down
    execute "DROP FUNCTION naturalsort(text);"
  end
end
```

PR #11, opened 2017-07-25
Add support for aggregate definitions

jlogsdon:
Hello! First, I'd like to apologize for coming out of nowhere with a "major" feature add. I'm developing a new system at work that will require some custom functions and aggregates and I desired a way to manage them through migrations. While this gem provided a framework to work with, it lacked two important features: handling functions with multiple signatures (#7, #10) and management of aggregate functions.

The implementation is directly tied to Postgres, but without the gem officially supporting more than one database I don't think there's a clean way around it (without a lot more work). If this a blocker I totally understand, the changes here are easy enough to patch in if we need.

If you're willing to accept this PR I'll work on README updates. I didn't want to spend too much time updating that if this will end up rejected.


jlogsdon:
I must be using a different rubocop version with my editor, most of those issues were not present when editing stuff. I'll go through and cleanup.

I should also note that I am starting to use this feature _right now_ so there may be some iteration after actually working with it in a real-world scenario.

teoljungberg:
First and foremost, thank you for submitting this PR. I didn't know of this feature, but after reading about it. I totally want it in this gem.

Some answers to your questions:

> The implementation is directly tied to Postgres

Postgres is the database I use on almost every Rails project I am on, and this gem is built around that premise. I've written the adapter so it could and would be interchanged by another gem such as `fx-mysql` or `fx-WHATEVER_DATABASE`. So that is totally fine.

Since this feature is new to me, and the diff is quite vast. It may take a while for me to review it and understand it. Just as a heads up.

teoljungberg:
And:

> First, I'd like to apologize for coming out of nowhere with a "major" feature add.

That is totally fine, in the future you can open an issue and we can discuss it before hand. But opening up a PR is fantastic too!

teoljungberg:
I thought about something, as this does not rely on any other refactorings in `F(x)` (besides excluding the aggregate functions from the normal functions list).

How would you feel about writing this as an extension to F(x) instead? That way, you are in more control of how it works, and we can see what changes are required in F(x) for these kinds of extensions to be made.

This is just a loose thought, I am unsure where I will land in the reasoning. But I wanted to lift the discussion.

What do you think?

jlogsdon:
> Since this feature is new to me, and the diff is quite vast. It may take a while for me to review it and understand it. Just as a heads up.

Completely understand!

> How would you feel about writing this as an extension to F(x) instead?

An interesting idea. I doubt I'll have time over the next 2 weeks (major feature deliverable followed by a week of travelling) to play around with an extension approach, but I don't see it being too difficult with something like this.

teoljungberg:
Hi @jlogsdon - are you still interesting in working on this feature?

jlogsdon:
Yes, things have just been a bit busier than expected. I have a few hours this weekend where I'll be looking at the extension approach. I'll report back here with my thoughts.

teoljungberg:
Pinging @jlogsdon.

reidab:
I'd also love to see this incorporated, either as an extension or as a core feature.

teoljungberg:
As this PR has been sitting for a while now, I'm going to close it.

I'll happily support having extensions to fx, but I don't think I'll consider this a core feature.

agustin:
Hi @teoljungberg I was looking into this and wondering if you already think about how to implement this as an extension? I'm happy to work on it if you are able to share some ideas already. I can imagine that we need to add an `Adapter` and extend some methods on the `postgres.rb` adapter, right? or do you imagine differently?

teoljungberg:
@agustin - I have not thought nor planned to work on an extension. Feel free to go ahead!

Use what is in this PR as a reference, and I’ll happily make fx’s code easier to work with for the sake of an extension. If you run into issues, do let me know.

agustin:
Just FYI I pushed [this new gem](https://github.com/agustin/fx-aggregate) as an extension to add support for aggregates. I based on this PR but also on the latest version of `fx`. There are some missing specs though, but it's working as expected (at least with the test cases and my personal use case)

PR #13, opened 2018-03-07
Strip shared leading whitespace from sql_definitions

naw:
If the whitespace isn't stripped, the whitespace finds its way into the actual function definitions. Successive runs of `schema:load` and `schema:dump` will add 6 spaces of indentation to the function definition (each time). The indentation keeps growing and growing, which results in a lot of churn in schema.rb and unnecessary merge conflicts between disparate source control branches.

### DIFF
--- a/lib/fx/statements/function.rb
+++ b/lib/fx/statements/function.rb
@@ -36,6 +36,7 @@ module Fx
             "version or sql_definition must be specified",
           )
         end
+        sql_definition = sql_definition.strip_heredoc if sql_definition
         sql_definition ||= Fx::Definition.new(name: name, version: version).to_sql
 
         Fx.database.create_function(sql_definition)
@@ -93,6 +94,7 @@ module Fx
           )
         end
 
+        sql_definition = sql_definition.strip_heredoc if sql_definition
         sql_definition ||= Fx::Definition.new(
           name: name,
           version: version,
--- a/lib/fx/statements/trigger.rb
+++ b/lib/fx/statements/trigger.rb
@@ -39,6 +39,7 @@ module Fx
           version = 1
         end
 
+        sql_definition = sql_definition.strip_heredoc if sql_definition
         sql_definition ||= Fx::Definition.new(
           name: name,
           version: version,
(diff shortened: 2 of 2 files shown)


teoljungberg:
Hi @naw 

Can you provide me with a diff of a schema before and after this change was applied?

naw:
@teoljungberg 

To be clear, this change does not actually alter any schemas retroactively. It prevents future unnecessary changes to schemas when you load and dump the schema.

Before this change, running `db:schema:load` followed by `db:migrate` will add 6 spaces of indentation to every function in schema.rb. Each successive run of `db:schema:load` followed by `db:migrate` will add 6 more spaces. You can see this in these two commits: https://github.com/naw/fx-indentation-example/commit/0a4f9d9c199b994d387eb2327fd715f5d05cdb40 and https://github.com/naw/fx-indentation-example/commit/052f76a8818b566928636a72cab59ba4371f741d

After this change, running `db:schema:load` followed by `db:migrate` will cause no extraneous schema changes.

So, you can't really compare a before/after diff, but you can compare what a repo would look like if you had run the commands with the path and without the patch in place:

Here, the patched version has no extra indentation after running schema commands, whereas the unpatched version has 12 extra spaces of indentation after running schema commands:

https://github.com/naw/fx-indentation-example/compare/0bd86...052f76

Also see the README here: https://github.com/naw/fx-indentation-example 

Note: I'm using `db:migrate` as an example of a commonly used command that dumps the schema. The issue isn't running migrations, but dumping the schema, so using `db:schema:dump` would cause the same issue. So, if you load your schema and then switch to a branch that has some new migrations in it and run the migrations, this will dump the schema and increase the indentation in the functions even though the migrations and the branch have nothing to do with those functions.

teoljungberg:
Thanks for the thorough explanation @naw!

I sat and thought about this, and wondered why [Scenic] doesn't have anything like this. But couldn't figure it out.

Either way - I'm fine with this approach. I'll merge and release a new version.

Thanks for your contribution!

[Scenic]: https://github.com/thoughtbot/scenic/

teoljungberg:
Released as of https://github.com/teoljungberg/fx/commit/cdb3180c3df1993e1cf52301fd577062290d8738

PR #17, opened 2018-12-14
Reverse function/trigger order in README

geoffharcourt:
The README describes how to create a trigger and a function, but when
building a trigger that relies on a function, the function needs to
exist first. Following the suggestions in order will create a migration
that doesn't work because of an uncreated function dependency.


teoljungberg:
Awesome, thanks Geoff!

PR #19, opened 2018-12-28
Split up `Trigger#definition` test

teoljungberg:
Newer PostgreSQL versions denote the table with its qualified name i.e
`public.users` over `users` which the older versions did.

This change relaxes the test to separately validate each part of the 
definition.

Changes:
- Update gemfiles. These changes were generated by `bundle exec
  appraisals install`.

  - Appraisals changed from 1.8 hash syntax to 1.9.
  - Add forgotten version bump of `fx`.
  - Update `bundle` version to the latest.

- Delete `.ruby-version`. As we support several ruby(1) versions, it
feels redundant to lock the project into using 2.3.*.


pjungwir:
Looks good to me! Thanks for taking that on. :-)

PR #25, opened 2019-01-05
Make `create_function` reversible with `version`.

pjungwir:
If you say `create_function("foo", version: 1)` then rolling back the
migration shouldn't fail, even though `drop_function` doesn't have a
`version` keyword argument.

Changes:

- Adds a test to verify that `create_function` is reversible even with a
  `version` parameter.
- Updates the command recorder to exclude `version` when calling
  `drop_function`. (I also considered just giving `drop_function` a new
  noop `version` parameter, which would keep the recorder code simpler,
  but it didn't seem worth polluting the signature of an important
  "public" method.)

### DIFF
--- a/lib/fx/command_recorder/function.rb
+++ b/lib/fx/command_recorder/function.rb
@@ -15,7 +15,7 @@ module Fx
       end
 
       def invert_create_function(args)
-        [:drop_function, args]
+        [:drop_function, [args[0], args[1].try(:except, :version)].compact]
       end
 
       def invert_drop_function(args)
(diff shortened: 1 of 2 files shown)


teoljungberg:
Hi @pjungwir, sorry for not responding here quicker.

`create_function("foo", version: 1)` should definitely be able to be rolled back. However, since version 1 is considered default when generating a new function you can work around the issue by omitting the `version:` argument.

This is also inline with how [Scenic](https://github.com/scenic-views/scenic) works.

In light of that, I'm closing this PR. Hope this answers your questions!

pjungwir:
I find this decision puzzling. Why would you want rollback to fail when the version is explicit but succeed when it's implicit? The explicitness seems like an accidental quality no one would expect to influence the behavior.

If you like I can submit a PR to scenic too. :-)

Issue #12, opened 2017-07-31
Replacing functions when updating, instead of dropping and recreating


jlogsdon:
As we've begun adding functions to our system we've found a minor but annoying pain point when updating functions. With the current behavior a call to `update_function` will first drop and then recreate the function. For simple use-cases this behavior works fine, but as your functions grow and begin to depend on each other this no longer works. The problem becomes worse with the addition of views through the scenic gem.

As an example, imagine this scenario:

* A function `age_years(dob date) returns int8` for calculating the number of years since `dob`.
* A function `age_bracket(dob date) returns text` returning a string of "child", "adolescent" or "adult" using `age_years`.
* A view (through scenic) that uses both `age_years(dob date)` and `age_bracket(dob date)`.

In the current system we have two options:

1. The migration will drop all dependencies, and then rebuild. This is essentially a non-starter in terms of complexity.
2. Use `CREATE OR REPLACE FUNCTION` in our function definitions, and in the migrations manually execute the correct up and down versions.

My current thinking is that adding a `replace_function` for migrations, with a generator flag `--replace`, would be enough to completely support these scenarios. Perhaps even make the default `replace_function` and keep the "tear it all down" `update_function` approach for when actually needed.

An alternative, that I don't like as much, would be changing `update_function` to "realize" when a script has `OR REPLACE` and skip dropping those functions. Although instead of a new `replace_function` method we could add the flag `replace: true` to this method instead.

teoljungberg:
Since `update_function` should be enough for the simple cases, I want to leave it as the default.

I like adding `replace_function`, it is the route scenic took. And they did not add a `--replace` flag, so I think we can opt out of adding that too and simply document it.

jlogsdon:
Sounds good to me! I'll start work on a PR to add `replace_function` soon.

reidab:
I'd love to see this too, since all my trigger handler functions fail to update because the triggers depend on them.

I've worked around it by using `CREATE OR REPLACE` and writing update migrations like this:

```ruby
class UpdateFunctionHandleCampgroundUpdateToVersion2 < ActiveRecord::Migration[5.0]
  def up
    create_function :handle_campground_update, version: 2
  end

  def down
    create_function :handle_campground_update, version: 1
  end
end
```

bbugh:
[...]

PR #26, opened 2019-04-24
Support functions with parameters

stoivo:
With postgres version above 10 you don't need to send the arguments for dropping function.

In our project we have created functions like car_active, it takes a date as an argument. When we needed to update it we would use the trick https://github.com/teoljungberg/fx/issues/12#issuecomment-390002371.

Since version 10 it is possible to drop functions without passing the arguments so I went ahead and implemented it today. 

I'm not 100% sure know you will take this. I found one of your old Pull request. I dont know why you closed it https://github.com/teoljungberg/fx/pull/8. 

I have seen there has been done some work in the past tring to implemnet both function with arguments and multiple functions same name with different agruments. This solved the first one functions with arguments. Multiple functions same name with different agruments is still unresolved and I  am fine with not haveing it.


stoivo:
This will not improve the experense on versions prior to 10 but for 10 and above it is better. postgres 9.6 will still be active untll 2021 November 11 https://www.postgresql.org/support/versioning/. My feeling are saying that a lot of servers and CI are using an older versions of postgres. ☹️

It says [here](https://docs.travis-ci.com/user/database-setup/#using-a-different-postgresql-version) that TravisCI by default is using 9.2 which is unsopported by postgres. 

Even will these downsides I would like to consider this patch since it is working good for 10 + which was released first on 2017 Oktober 5 and does not ham any thing(execpt complexety).

PR #27, opened 2019-04-25
Support functions with parameters

stoivo:
copy of #26 just a different sourcebranch

With postgres version above 10 you don't need to send the arguments for dropping function.

In our project we have created functions like car_active, it takes a date as an argument. When we needed to update it we would use the trick https://github.com/teoljungberg/fx/issues/12#issuecomment-390002371.

Since version 10 it is possible to drop functions without passing the arguments so I went ahead and implemented it today. 

I'm not 100% sure know you will take this. I found one of your old Pull request. I dont know why you closed it https://github.com/teoljungberg/fx/pull/8. 

I have seen there has been done some work in the past tring to implemnet both function with arguments and multiple functions same name with different agruments. This solved the first one functions with arguments. Multiple functions same name with different agruments is still unresolved and I  am fine with not haveing it.

This will not improve the experense on versions prior to 10 but for 10 and above it is better. postgres 9.6 will still be active untll 2021 November 11 https://www.postgresql.org/support/versioning/. My feeling are saying that a lot of servers and CI are using an older versions of postgres. ☹️

It says [here](https://docs.travis-ci.com/user/database-setup/#using-a-different-postgresql-version) that TravisCI by default is using 9.2 which is unsopported by postgres. 

Even will these downsides I would like to consider this patch since it is working good for 10 + which was released first on 2017 Oktober 5 and does not ham any thing(execpt complexety).

### DIFF
--- a/spec/acceptance/user_manages_functions_spec.rb
+++ b/spec/acceptance/user_manages_functions_spec.rb
@@ -48,6 +48,7 @@ describe "User manages functions" do
     successfully "rake db:migrate"
 
     result = execute("SELECT * FROM adder(1, 2) AS result")
+    result["result"] = result["result"].to_i
     expect(result).to eq("result" => 3)
 
     successfully "rails destroy fx:function adder"
--- a/spec/acceptance/user_manages_functions_spec.rb
+++ b/spec/acceptance/user_manages_functions_spec.rb
@@ -34,4 +34,23 @@ describe "User manages functions" do
     result = execute("SELECT * FROM test() AS result")
     expect(result).to eq("result" => "testest")
   end
+
+  it "handles functions with arguments" do
+    successfully "rails generate fx:function adder"
+    write_function_definition "adder_v01", <<-EOS
+      CREATE FUNCTION adder(x int, y int)
+      RETURNS int AS $$
+      BEGIN
+          RETURN $1 + $2;
+      END;
+      $$ LANGUAGE plpgsql;
+    EOS
+    successfully "rake db:migrate"
+
+    result = execute("SELECT * FROM adder(1, 2) AS result")
+    expect(result).to eq("result" => 3)
+
+    successfully "rails destroy fx:function adder"
+    successfully "rake db:migrate"
+  end
 end
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,17 @@
 addons:
-  postgresql: "9.4"
+  postgresql: "10"
+  apt:
+    packages:
+    - postgresql-10
+    - postgresql-client-10
 before_install:
   - "echo '--colour' > ~/.rspec"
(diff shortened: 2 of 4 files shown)


spec/acceptance/user_manages_functions_spec.rb
```diff
@@ -34,4 +34,36 @@
...
+      RETURNS int AS $$
+      BEGIN
+          RETURN $1 + $2;
+      END;
+      $$ LANGUAGE plpgsql;
+    EOS
+    successfully "rake db:migrate"
+
+    result = execute("SELECT * FROM adder(1, 2) AS result")
+    # require "byebug"
```
teoljungberg: Could you remove these debug statements and reformat the test file to follow how the test suite is formatted?

spec/acceptance/user_manages_functions_spec.rb
```diff
@@ -34,4 +34,36 @@
...
+    # byebug
+    puts "PG::VERSION: #{PG::VERSION.inspect}"
+    puts "ActiveRecord.gem_version: #{ActiveRecord.version.inspect}"
+    puts "result: #{result.inspect}"
+
+
+    # if PG::VERSION >= "0.18.0"
+    if ActiveRecord.version > Gem::Version.new("5.0.0")
+      expect(result).to eq("result" => 3)
```
teoljungberg: You can omit the conditional by enforcing the result to be either a string or an integer. I'd prefer that.

spec/acceptance/user_manages_functions_spec.rb
stoivo: I removed the debug statments. Not sure what you ment with reformating. The spec is copied from https://github.com/teoljungberg/fx/pull/8/files

stoivo:
Thank you for mergeing the other PRs I will rebase these 2 and check that the specs are passing.

stoivo:
ping

teoljungberg:
I'll get to this on Friday. Thanks for rebasing it

teoljungberg:
I'm cutting a maintenance release 0.5.0 that includes your previous work in #28 and #31.

teoljungberg:
I'm going to leave this unreleased for a while, please test it out in the meantime by pointing the gem towards master.

stoivo:
Will do and report any issues. Thanks for taking the time to review and merger

schnmudgal:
Can we release this now as a version (instead of master reference)?

teoljungberg:
@schnmudgal Yup, I'm lagging behind cutting a release for this. I want to find time and energy to do it within the next 1-2 weeks as it's been a real long time since I released anything.

dnamsons:
Hey @teoljungberg, I was wondering if you have an estimate for a release for this fix?

PR #29, opened 2019-04-30
Remove gemfile.lock so it will try with the lastes releases

stoivo:
Per Appraisal recommendations:
		When using Appraisal, we recommend you check in the Gemfiles that Appraisal
		generates within the gemfiles directory, but exclude the lockfiles
		there (*.gemfile.lock.)

https://github.com/thoughtbot/appraisal#version-control


teoljungberg:
Since the branch was out of sync, I merged your work as 392c434efb5edebf0304e73f74b30f3ffcf3eacc and retain the authorship to you.

PR #35, opened 2019-06-15
Remove 'public' schema name from the function dump

mri-dula:
Thank you for this very useful gem! I have made two modifications as listed below:

A] When dumping the functions to the `schema.rb` file, the schema name `public` gets used in the `CREATE FUNCTION` command. For example: `CREATE OR REPLACE FUNCTION public.adder() ....`

When using with the [Apartment gem](https://github.com/influitive/apartment), this is a slight issue. If you are unaware of this Gem, it is a very useful gem used for implementing multi-tenant web apps. The `schema.rb` is run multiple times by this gem, and the contents are created for each of the tenants of the app. For example, if the app has 2 tenants - `client1`, `client2`, there will be three `schema`s in the Postgres database - `public`, `client1` and `client2` all with the same tables and views.

However, because the `public` schema name gets used, the functions get created only in the `public` schema and not in the tenant` schemas. I have fixed this.

B] When I print `PG.connect.server_version`, it prints **90613**. In my version, if there is a function signature `adder(int x, int y)` it does not get dropped by either `DROP FUNCTION adder;` OR `DROP FUNCTION adder();`. Only `DROP FUNCTION adder(int, int)` works. I have fixed this.

### DIFF
--- a/lib/fx/adapters/postgres.rb
+++ b/lib/fx/adapters/postgres.rb
@@ -26,7 +26,7 @@ module Fx
       
       # The SQL query used by F(x) to retrieve the function name 
       # with its signature.
-      FUNCTION_NAME_WITH_SIGNATURE = <<-EOS.freeze
+      FUNCTION_NAMES_WITH_THEIR_SIGNATURE = <<-EOS.freeze
           SELECT
               pp.proname AS name,
               CONCAT(pp.proname, '(', pg_get_function_identity_arguments(pp.oid), ')') AS signature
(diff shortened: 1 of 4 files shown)


maheshm:
Probably you want to have two different PRs for each of these fixes and raise bugs for each before raising the fix.

For part A, AFAIK you should be handling it in your sql as you can see in the spec. If you do not have the public.* in the sql file then it should not be there in the schema dump either. Can you create a bug and move this discussion there?

Part B - I just experienced the same issue. Your solution is probably in right direction, but if there are way too many functions this is going to create a huge array in memory. That should be still ok, but we can achieve the same by using the query directly which is much more efficient.

lib/fx/adapters/postgres.rb
```diff
@@ -128,7 +144,8 @@ def drop_function(name)
         if support_drop_function_without_args
           execute "DROP FUNCTION #{name};"
         else
-          execute "DROP FUNCTION #{name}();"
+          signature = function_signature_by_name name
+          execute "DROP FUNCTION #{signature};"
```
maheshm: You can get inspiration from this:
https://dba.stackexchange.com/questions/122742/how-to-drop-all-of-my-functions-in-postgresql

maheshm:
BTW, https://github.com/teoljungberg/fx/pull/10 is already solving part B. Please do check that so that work is not duplicated.

mri-dula:
@maheshm, Thanks for your comments. I will create separate bugs and create a pull request for each. I will also check #10 out.

PR #36, opened 2019-06-19
use the decorated connection to check server version

lastobelus:
otherwise fails on cloud providers not using a standard local postgres socket

### DIFF
--- a/lib/fx/version.rb
+++ b/lib/fx/version.rb
@@ -1,4 +1,4 @@
 module Fx
   # @api private
-  VERSION = "0.5.0"
+  VERSION = "0.5.1"
 end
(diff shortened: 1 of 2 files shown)


maheshm:
Fixes #37

maheshm:
Can you check why CI is failing?

teoljungberg:
They all fail due to the same error.

```
PG::ConnectionBad:
       could not connect to server: No such file or directory
       	Is the server running locally and accepting
       	connections on Unix domain socket "/var/run/postgresql/.s.PGSQL.5433"?
```

teoljungberg:
It is fixed with the latest master now.

lib/fx/version.rb
```diff
@@ -1,4 +1,4 @@
 module Fx
   # @api private
-  VERSION = "0.5.0"
+  VERSION = "0.5.1"
```
teoljungberg: Please don't bump the version - I'll do that when it's time for release.

PR #48, opened 2019-09-02
Handle schema-qualified function names

gumchum:
Updates the function generator and the `Fx::Definition` so that schema-qualified function names can be specified.


teoljungberg:
I can't see if any tests have run for this. Your repository seems to have been deleted?

image

PR #49, opened 2019-09-17
Use db connection provided by Rails

rylwin:
First off, just want to say thanks for creating this gem -- I'm excited to use it in my project!

I might be strange, but on my dev machine I don't have a postgres database for my user (i.e., if I just type `psql` at the console, I get an error "psql: FATAL:  database "ryan" does not exist"). I have to specify the dbname. Because of this, I get an error when the code attempt to check the postgres server version. 

I wasn't sure what a good way to build a test for this would be without impacting the dev machine (I don't want to drop anyone's DB!), so I did not include an update to any spec file. However, before the change, specs fail for me. With the change, I'm green.

Looking forward to any comments/feedback on this change. Thanks again and commit body below:

------

Using the Rails db connection ensures we use valid database connection
information when checking the postgres server version.

Without specifying credentials, `PG` attempts to connect to the default
database (i.e., the default one for the current user) which, if it does
not exist, causes an error:

```
  Failure/Error: PG.connect.server_version >= 10_00_00

  PG::ConnectionBad:
    FATAL:  database "xxxxx" does not exist
```

By accessing the raw connection through `@connectable`, we avoid this
issue because we end up using the database configuration as specified by
the Rails app.

### DIFF
--- a/lib/fx/adapters/postgres.rb
+++ b/lib/fx/adapters/postgres.rb
@@ -159,7 +159,8 @@ module Fx
         # https://www.postgresql.org/docs/9.6/sql-dropfunction.html
         # https://www.postgresql.org/docs/10/sql-dropfunction.html
 
-        PG.connect.server_version >= 10_00_00
+        pg_connection = connectable.connection.raw_connection
+        pg_connection.server_version >= 10_00_00
       end
     end
   end


teoljungberg:
Sorry for taking so long to loop back, life's been hectic.

I'm very much in favor of this, how can I test so things still work as they should? Could you summarize some reproduction steps?

rylwin:
The issue occurs when a database matching the current "$USER" name does not exist. This happens because `PG.connect` attempts to connect to db named "$USER", which is the same behavior as if you were to run `psql` in your terminal. (Note that the issue would also occur if any other connection details do not match the postgres defaults.)

To reproduce the issue I'm describing, drop your user's db with `dropdb "$USER"`. Then run the test suite on `master`. You'll see several failures. The failure message you're looking for will be similar to the following:

```
     Failure/Error: PG.connect.server_version >= 10_00_00 
     
     PG::ConnectionBad:
       FATAL:  database "ryan" does not exist
     # ./lib/fx/adapters/postgres.rb:162:in `support_drop_function_without_args'
     # ./lib/fx/adapters/postgres.rb:128:in `drop_function'
     # ./spec/fx/adapters/postgres_spec.rb:70:in `block (4 levels) in <module:Adapters>'
     # ./spec/acceptance_helper.rb:8:in `block (3 levels) in <top (required)>'
     # ./spec/acceptance_helper.rb:7:in `chdir'
     # ./spec/acceptance_helper.rb:7:in `block (2 levels) in <top (required)>'
     # ./spec/spec_helper.rb:13:in `block (2 levels) in <top (required)>'
```

Now if you checkout this PR and rerun the test suite, you should be green (or at least any failures should be unrelated to the change in `#support_drop_function_without_args`.

teoljungberg:
Many thanks @rylwin (and @sobrinho for the nudge)!

Issue #34, opened 2019-06-13
Schema dump fails in postgis


maheshm:
https://github.com/teoljungberg/fx/blob/53f8b3fb651e7574f98f6ff3685361a13de0981c/lib/fx/adapters/postgres/functions.rb#L11

This throws the following error if you have aggregate functions (Postgis has a lot of these):

```
ActiveRecord::StatementInvalid: PG::WrongObjectType: ERROR:  "array_accum" is an aggregate function
:           SELECT
              pp.proname AS name,
              pg_get_functiondef(pp.oid) AS definition
          FROM pg_proc pp
          JOIN pg_namespace pn
              ON pn.oid = pp.pronamespace
          LEFT JOIN pg_depend pd
              ON pd.objid = pp.oid AND pd.deptype = 'e'
          WHERE pn.nspname = 'public' AND pd.objid IS NULL
          ORDER BY pp.oid;
```

Stacktrace if that helps:

```
/bundle/gems/activerecord-5.1.6.2/lib/active_record/connection_adapters/postgresql/database_statements.rb:73:in `async_exec'
/bundle/gems/activerecord-5.1.6.2/lib/active_record/connection_adapters/postgresql/database_statements.rb:73:in `block (2 levels) in execute'
/bundle/gems/activesupport-5.1.6.2/lib/active_support/dependencies/interlock.rb:46:in `block in permit_concurrent_loads'
/bundle/gems/activesupport-5.1.6.2/lib/active_support/concurrency/share_lock.rb:185:in `yield_shares'
/bundle/gems/activesupport-5.1.6.2/lib/active_support/dependencies/interlock.rb:45:in `permit_concurrent_loads'
/bundle/gems/activerecord-5.1.6.2/lib/active_record/connection_adapters/postgresql/database_statements.rb:72:in `block in execute'
/bundle/gems/activerecord-5.1.6.2/lib/active_record/connection_adapters/abstract_adapter.rb:613:in `block (2 levels) in log'
/bundle/gems/activerecord-5.1.6.2/lib/active_record/connection_adapters/abstract_adapter.rb:612:in `block in log'
/bundle/gems/activesupport-5.1.6.2/lib/active_support/notifications/instrumenter.rb:21:in `instrument'
/bundle/gems/activerecord-5.1.6.2/lib/active_record/connection_adapters/abstract_adapter.rb:604:in `log'
[...]

PR #38, opened 2019-08-19
Skip aggregate functions on schema dumping (#34)

vladimirtemnikov:
This should resolve #34 
Postgres can't return definitions for aggregate functions, so we need to exclude them from query. 

In addition the way to do this varies by Postgres version.
Accordingly to PG 11 changelog:

> Replace system catalog pg_proc's proisagg and proiswindow columns with prokind (Peter Eisentraut)
> This new column more clearly distinguishes functions, procedures, aggregates, and window functions.

it's possible by adding condition `pg_proc.prokind != 'a'` for Postgres >= 11 version or `pg_proc.proisagg = 'f'` for other versions.

PR #50, opened 2019-11-29
Does not include aggregates when dumping schema (#34)

johydoski:
Calling `pg_get_functiondef(some_oid)` on aggregates fails. Assuming
this gem is not meant to support aggregate functions, aggregate functions
are now excluded via omitting them from the query that pulls function
definitions from PG. This is achieved by checking for the function's
presence in table `pg_aggregate`.

`pg_aggregate` is present in all officially supported PG versions ( >= 9.4 ).

### DIFF
--- a/lib/fx/adapters/postgres/functions.rb
+++ b/lib/fx/adapters/postgres/functions.rb
@@ -17,7 +17,10 @@ module Fx
               ON pn.oid = pp.pronamespace
           LEFT JOIN pg_depend pd
               ON pd.objid = pp.oid AND pd.deptype = 'e'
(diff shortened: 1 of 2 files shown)


vladimirtemnikov:
@teoljungberg This looks prettier than https://github.com/teoljungberg/fx/pull/38/files

bradrobertson:
I can confirm this works for us on PG 11 and PG 10

bradrobertson:
Can we get some movement on this? The implementation seems pretty straightforward and tested. This would be a blocker for anyone needing `postgis` with `fx`

PR #51, opened 2020-02-22
Fix outputting escapes in schema output

rylwin:
Update `Function#to_schema` in order to generate schema output that
maintains escapes appropriately.

For example, say you have a function that includes an escape sequence,
such as the popular `naturalsort` function (source
http://www.rhodiumtoad.org.uk/junk/naturalsort.sql):

```
create or replace function naturalsort(text)
  returns bytea
  language sql
  immutable strict
as $f$
  select string_agg(convert_to(coalesce(r[2],
      length(length(r[1])::text) || length(r[1])::text || r[1]),
      'SQL_ASCII'),'\x00')
    from regexp_matches($1, '0*([0-9]+)|([^0-9]+)', 'g') r;
$f$;
```

Note the null byte `\x00`. Simply printing the function definition to
the schema causes issues as, when a user performs `rails
db:schema:load`, Rails complains: "ArgumentError: string contains null
byte."

In order to prevent Rails from interpreting the string as having a null
byte in the middle of the string, resulting in an `ArgumentError`, we
escape the null byte. This results in Rails being able to load the
string to and to create the function appropriately.

I further tested this change by directly checking Postgres' definition
of the function with `\df+ {functionname}`.

### DIFF
--- a/lib/fx/function.rb
+++ b/lib/fx/function.rb
@@ -18,7 +18,7 @@ module Fx
     def to_schema
       <<-SCHEMA.indent(2)
 create_function :#{name}, sql_definition: <<-\SQL
-#{definition.indent(4).rstrip}
+#{definition.indent(4).rstrip.gsub("\\", "\\\\\\\\")}
 SQL
       SCHEMA
     end
--- a/lib/fx/function.rb
+++ b/lib/fx/function.rb
@@ -2,6 +2,9 @@ module Fx
   # @api private
   class Function
     include Comparable
+    ESCAPE_MAP = {
+      '\\' => '\\\\',
+    }.freeze
 
     attr_reader :name, :definition
     delegate :<=>, to: :name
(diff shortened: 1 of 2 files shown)


lib/fx/function.rb
```diff
@@ -18,7 +18,7 @@ def ==(other)
     def to_schema
       <<-SCHEMA.indent(2)
 create_function :#{name}, sql_definition: <<-\SQL
-#{definition.indent(4).rstrip}
+#{definition.indent(4).rstrip.gsub("\\", "\\\\\\\\")}
```
teoljungberg, line 21: Something about this tells me that this will be error prone.

Is there a generic escape sequence that Rails or any other database related library uses we can reuse here instead?

rylwin: Good push @teoljungberg. The best example I found is from https://github.com/rails/rails/blob/7a321d8429e3d7b9d8aa97a44ce4a4b6cb62de70/actionview/lib/action_view/helpers/javascript_helper.rb#L34

It takes advantage of `gsub`'s ability to accept a hash as the second argument, wherein each match is replaced by the value found by looking up the match in the hash (TIL, https://ruby-doc.org/core-2.7.1/String.html#method-i-gsub) 

What do you think about this implementation?

teoljungberg:
#71

teoljungberg:
I think #71 resolves this at least.

PR #52, opened 2020-04-03
Add support to replace functions with dependencies

sobrinho:
This will add support to update functions without dropping and recreating dependencies.

### DIFF
--- a/spec/fx/adapters/postgres_spec.rb
+++ b/spec/fx/adapters/postgres_spec.rb
@@ -20,6 +20,50 @@ module Fx::Adapters
       end
     end
 
+    describe "#update_function" do
+      it "successfully updates a function" do
+        allow(adapter).to receive(:drop_function)
+        allow(adapter).to receive(:create_function)
+
+        name = "test"
+        sql_definition = <<-EOS
+          CREATE OR REPLACE FUNCTION test()
+          RETURNS text AS $$
+          BEGIN
+              RETURN 'test';
+          END;
+          $$ LANGUAGE plpgsql;
+        EOS
+
+        adapter = Postgres.new
+        adapter.update_function(name, sql_definition)
+
+        expect(adapter).to have_received(:drop_function).with(name)
+        expect(adapter).to have_received(:create_function).with(sql_definition)
+      end
+
+      it "successfully replaces a function" do
+        allow(adapter).to receive(:drop_function)
+        allow(adapter).to receive(:create_function)
+
+        name = "test"
+        sql_definition = <<-EOS
+          CREATE OR REPLACE FUNCTION test()
+          RETURNS text AS $$
+          BEGIN
+              RETURN 'test';
+          END;
+          $$ LANGUAGE plpgsql;
+        EOS
+
+        adapter = Postgres.new
+        adapter.update_function(name, sql_definition, replace: true)
+
+        expect(adapter).not_to have_received(:drop_function)
+        expect(adapter).to have_received(:create_function).with(sql_definition)
+      end
+    end
+
     describe "#create_trigger" do
       it "successfully creates a trigger" do
         connection.execute <<-EOS
--- a/README.md
+++ b/README.md
@@ -75,6 +75,21 @@ version 1, created a copy of that definition as version 2, and created a
 migration to update to the version 2 schema. All that's left for you to do is
 tweak the schema in the new definition and run the `update_function` migration.
 
+## What about updating functions with existing triggers?
+
+If you try to update a function that already have dependencies such as triggers,
+you must use the replace flag:
+
+```sh
+% rails generate fx:function uppercase_users_name --replace
+      create  db/functions/uppercase_users_name_v02.sql
+      create  db/migrate/[TIMESTAMP]_update_function_uppercase_users_name_to_version_2.rb
+```
+
+While the default behavior allows you to change the function signature, if you
+don't change the function signature, you can replace the existing function
+without recreating its dependencies.
+
 ## I don't need this trigger or function anymore. Make it go away.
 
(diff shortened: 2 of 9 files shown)


sobrinho:
@teoljungberg in other words, today we have to do this:

```ruby
def up
  drop_trigger :blah, on: :blah
  update_function :blah, version: 2
  create_trigger :blah, on: :blah
end

def down
  drop_trigger :blah, on: :blah
  update_function :blah, version: 1
  create_trigger :blah, on: :blah
end
```

And after the PR, we just do:

```ruby
def change
  update_function :blah, version: 2, revert_to_version: 1, replace: true
end
```

spec/fx/adapters/postgres_spec.rb
```diff
@@ -20,6 +20,38 @@ module Fx::Adapters
       end
     end
 
+    describe "#update_function" do
+      let(:name) { "test" }
```
teoljungberg, line 24: I don't really rely on `let` anywhere else in the codebase, can you instead inline the 'let's as local variables?

spec/fx/adapters/postgres_spec.rb
```diff
@@ -20,6 +20,38 @@ module Fx::Adapters
...
+              RETURN 'test';
+          END;
+          $$ LANGUAGE plpgsql;
+        EOS
+      end
+
+      it "successfully updates a function" do
+        adapter = Postgres.new
+
+        expect(adapter).to receive(:drop_function).with(name)
```
teoljungberg, line 39: Can you split this up into an `allow` step and then `expect`? Look elsewhere in the application for how I normally like to do that.

spec/fx/adapters/postgres_spec.rb
sobrinho: The idea behind `let` is to avoid duplicating things across different scenarios.

Even so do you still want to change to local variables?

spec/fx/adapters/postgres_spec.rb
sobrinho: This is also about duplication:

```ruby
        adapter = Postgres.new

        expect(adapter).not_to receive(:drop_function)
        expect(adapter).to receive(:create_function).with(sql_definition)

        adapter.update_function(name, sql_definition, replace: true)
```

VS:

```ruby
        allow(adapter).not_to receive(:drop_function)
        allow(adapter).to receive(:create_function).with(sql_definition)

        adapter = Postgres.new
        adapter.update_function(name, sql_definition, replace: true)

        expect(adapter).not_to have_received(:drop_function)
        expect(adapter).to have_received(:create_function).with(sql_definition)
```

spec/fx/adapters/postgres_spec.rb
teoljungberg: That's right.

spec/fx/adapters/postgres_spec.rb
teoljungberg: I want it duplicated to keep each test more readable

teoljungberg: Try to follow the style of the tests already in place

bf4:
I was just hacking the gem for the feature in this PR.  I'm happy to take it over if OP @sobrinho is taking a break.

Basic impl looks good to me

sobrinho:
I will do the requested changes in a couple of hours :)

bf4:
I'm happy to take this over if it's gonna be stale much longer. I've added some work arounds in my app, but I'd like to use it :)

teoljungberg:
Please do @bf4. I'd like this feature too.

sobrinho:
@teoljungberg would you review it again?

teoljungberg:
As we can do this today, in a bit of a clunkier fashion, I think it's fine and we can leave this as is.

sobrinho:
@teoljungberg this is a small effort in terms of code and maintenance in exchange of a big convenience for heavy users. Would you reconsider it?

teoljungberg:
I'd love to hear from others to reconsider it, and as I said

> As we can do this today, in a bit of a clunkier fashion, I think it's fine and we can leave this as is.

I don't see this as something that needs to be considered right now. Without having more input.

bf4:
@teoljungberg for me, I've worked around this by writing my own update functions as described in #63 

```ruby
  # Could be simplified after https://github.com/teoljungberg/fx/pull/52 is merged
  def db_update_function(function_name, version:, revert_to_version:, replace: true)
    if replace
      reversible do |dir|
        dir.up do
          safely_execute function_sql(function_name, version: version)
        end
        dir.down do
          safely_execute function_sql(function_name, version: revert_to_version)
        end
      end
    else
      safety_assured do
        update_function function_name version: version, revert_to_version: revert_to_version
      end
    end
  end
```

if this PR were merged, even as an opt-in (rather than opt-out), I could change my code to

```ruby
  def db_update_function(function_name, version:, revert_to_version:, replace: true)
    safety_assured do
      update_function function_name version: version, revert_to_version: revert_to_version, replace: replace
    end
  end
```

because we use `CREATE OR REPLACE FUNCTION` for all our function files

teoljungberg:
Thanks for the additional input @bf4 - good to have for future context.

trcarden:
> I'd love to hear from others to reconsider it, and as I said
> 
> > As we can do this today, in a bit of a clunkier fashion, I think it's fine and we can leave this as is.
> 
> I don't see this as something that needs to be considered right now. Without having more input.

@teoljungberg here is a bit more input:

- We use the functions above in db views (for various reasons)
- Dropping a function would require coordination across all the views that depend on it 
- Adding replacement support so we don't have to drop and recreate views (which can be expensive, especially if they are materialized and large) would be a win
- This PR, while originally built to support triggers, could also be used for all other dependent object types

PR #54, opened 2020-04-21
Drop EOL Rails'

sobrinho:
Closes #44


sobrinho:
BUT Ruby 2.2, 2.3 and 2.4 is EOL now.

Maybe we want to drop it as well?

teoljungberg:
I'd break this PR out into multiple:

1) that removes the EOL versions of Ruby / Rails
2) that adds new version of Ruby / Rails that are needed

In which order they're done is less of a concern. But I like to keep the history clean.

teoljungberg:
I'll investigate why Travis is acting up, I can't sign in so it might be a fluke somewhere causing this.

sobrinho:
@teoljungberg done, changed only to remove the EOL Rails'.

After this one being merged, I will do another one handling ruby versions.

sobrinho:
In short, after this PR, we will support only Ruby `>= 2.2.2`, it's the least version Rails 5.2 supports.

teoljungberg:
Can you try to close this PR and open up a new one and see if that kicks Travis back online?

sobrinho:
Sure, let me try.

Issue #55, opened 2020-04-21
Is option `on` for #create_trigger useless?


QNester:
I looked at `#create_trigger` method code and noticed than option `on` not used in code. I tried run my migration without this option and migration successful done.

Is option `on` for `#create_trigger` useless? What do you think about using this option to interpolate sql definition?

For example: 
```
CREATE TRIGGER update_state_changed_at
    BEFORE UPDATE ON %{on}
    FOR EACH ROW
    WHEN (OLD.state <> NEW.state)
EXECUTE PROCEDURE update_state_changed_at();
```

I think this is easy and I can make pull request if you don't mind

teoljungberg:
If it's not used - it's something we should remove. It's probably a relic of an older spike on the implementation.

QNester:
But what about add feature with variables in triggers? 

One trigger file can be reused multiple times for  different tables. I think it can be useful and flexible.

teoljungberg:
Sure - we can do that. But that seems to make more sense as a feature addition rather than making an unused feature into one.

QNester:
Yes, I agree. I'll try make PR this week if you don't mind

teoljungberg:
Have at it!

QNester:
Can you check it please? This PR blocks my production development :(

teoljungberg:
What PR? #57?

I thought you were going to remove the `on` option, as that is what this issue is about.

QNester:
Excuse me, I read issue description again and now I see that it is really confusing. As I said - my English is not good :(

I wanted **use** `on` option to parametrize trigger, and wrote about it in issue: 
> What do you think about using this option to interpolate sql definition?

But you said that bad idea to use legacy option for it (and I agreed), and I made PR #57 with new options for interpolate sql definition. I think that idea to **remove** `on` option can be so problem, because gem users can run into problem with backward compatibility.

bf4:
I thought that 'on' could be useful for organizational purposes, e.g. db/triggers/${table_name}/${trigger_name}.sql but otherwise, perhaps we should make a PR to remove it since it's not used?

teoljungberg:
I'd be up for removing unused options, if such as PR arises.

PR #57, opened 2020-04-23
Support `table_name` variable in triggers sql definition

QNester:
### Parametrized trigger support for #55 

After this PR users can use parametrized triggers. For example:
```ruby
#migration
class CreateTriggersUpdateStateChangedAt < ActiveRecord::Migration[5.0]
  def change
    create_trigger :update_state_changed_at, params: { table_name: :users }
  end
end
```

```sql
# trigger 
CREATE TRIGGER update_state_changed_at
    BEFORE UPDATE ON %{table_name}
    FOR EACH ROW
    WHEN (OLD.state <> NEW.state)
EXECUTE PROCEDURE update_state_changed_at();
```

I think that **you should remove legacy `on` option in another** PR if you want. I think that removing this option can bad affect for current users of gem.

Sorry for not adding instructions to readme. I have no enough English for it :(

### DIFF
--- a/lib/fx/statements/trigger.rb
+++ b/lib/fx/statements/trigger.rb
@@ -46,6 +46,16 @@ module Fx
           type: DEFINTION_TYPE,
         ).to_sql
 
+        begin
+          interpolate_params = on ? { table_name: on } : {}
+          sql_definition = sql_definition % interpolate_params if sql_definition
+        rescue KeyError
+          raise(
+            ArgumentError,
+            "`on` option required for sql definition with `table_name` variable"
+          )
+        end
+
         Fx.database.create_trigger(sql_definition)
       end
 
@@ -123,6 +133,8 @@ module Fx
           version: version,
           type: DEFINTION_TYPE,
         ).to_sql
+
+        sql_definition = sql_definition % { table_name: on } if sql_definition && on
 
         Fx.database.update_trigger(
           name,
--- a/spec/fx/statements/trigger_spec.rb
+++ b/spec/fx/statements/trigger_spec.rb
@@ -41,6 +41,26 @@ describe Fx::Statements::Trigger, :db do
         /cannot both be set/,
       )
     end
+
+    it 'interpolates table_name in definition with `on` option' do
+      database = stubbed_database
+      definition = parametrized_definition
+      table_name = 'test_table'
+
+      connection.create_trigger(:test, on: table_name)
+
+      expect(database).to have_received(:create_trigger).
+        with(definition.to_sql % { table_name: table_name })
+      expect(Fx::Definition).to have_received(:new).
+        with(name: :test, version: 1, type: "trigger")
+    end
+
+    it 'raise error `ArgumentError` if definition expect table_name and `on` was not passed', focus: true do
+      stubbed_database
+      parametrized_definition
+
+      expect { connection.create_trigger(:test) }.to raise_error(ArgumentError)
+    end
   end
 
   describe "#drop_trigger" do
@@ -116,6 +136,25 @@ describe Fx::Statements::Trigger, :db do
         /cannot both be set/,
       )
     end
+
+    it 'interpolates table_name in definition with `on` option' do
+      database = stubbed_database
+      definition = parametrized_definition
+      table_name = :users
+
+      connection.update_trigger(:test, on: table_name, version: 3)
+
+      expect(database).to have_received(:update_trigger).with(
+        :test,
+        on: :users,
+        sql_definition: definition.to_sql % { table_name: table_name },
+      )
+      expect(Fx::Definition).to have_received(:new).with(
+        name: :test,
+        version: 3,
+        type: "trigger",
+      )
+    end
   end
 
   def stubbed_database
@@ -126,6 +165,12 @@ describe Fx::Statements::Trigger, :db do
 
   def stubbed_definition
     instance_double("Fx::Definition", to_sql: nil).tap do |stubbed_definition|
+      allow(Fx::Definition).to receive(:new).and_return(stubbed_definition)
+    end
+  end
+
+  def parametrized_definition
+    instance_double("Fx::Definition", to_sql: 'parametrized %{table_name}').tap do |stubbed_definition|
       allow(Fx::Definition).to receive(:new).and_return(stubbed_definition)
     end
   end


teoljungberg:
Thanks for the PR!

This solves half the issue, the other side of the coin is to be able to dump these triggers and functions into the schema.

teoljungberg:
I would like this pr to solve the functions side of the problem. As well as being able to dump the schema for both functions and triggers with parameters

QNester:
As I can see, triggers are dumping  into schema fine in current code. For example: 
1. I created two models: `User` and `Admin` with two columns - `state` and `state_changed_at`
2. I created function with followed SQL:
```sql
CREATE OR REPLACE FUNCTION update_state_changed_at() RETURNS TRIGGER AS $$
BEGIN
    NEW.state_changed_at = now();
    RETURN NEW;
END;
$$ LANGUAGE plpgsql;
```
Migration for function:
```ruby
class CreateFunctionUpdateStateChangedAt < ActiveRecord::Migration[5.0]
  def change
    create_function :update_state_changed_at
  end
end
```
3. I created triggers with following SQL:
```sql
CREATE TRIGGER update_state_changed_at
    BEFORE UPDATE ON %{table}
    FOR EACH ROW
    WHEN (OLD.state <> NEW.state)
EXECUTE PROCEDURE update_state_changed_at();
```
and following migration:
```ruby
class CreateTriggerUpdateStateChangedAt < ActiveRecord::Migration[5.0]
  def change
    [User, Admin].each do |model|
      create_trigger :update_state_changed_at, params: { table: model.table_name }
    end
  end
end
```
4. After `rails db:migrate` I see in schema:
```ruby
ActiveRecord::Schema.define(version: 2020_05_01_091752) do

  # These are extensions that must be enabled in order to support this database
  enable_extension "plpgsql"

  create_table "admins", force: :cascade do |t|
    t.string "state"
    t.datetime "state_changed_at"
    t.datetime "created_at", precision: 6, null: false
    t.datetime "updated_at", precision: 6, null: false
  end

  create_table "users", force: :cascade do |t|
    t.string "state"
    t.datetime "state_changed_at"
    t.datetime "created_at", precision: 6, null: false
    t.datetime "updated_at", precision: 6, null: false
  end


  create_function :update_state_changed_at, sql_definition: <<-SQL
      CREATE OR REPLACE FUNCTION public.update_state_changed_at()
       RETURNS trigger
       LANGUAGE plpgsql
      AS $function$
      BEGIN
          NEW.state_changed_at = now();
          RETURN NEW;
      END;
      $function$
  SQL

  create_trigger :update_state_changed_at, sql_definition: <<-SQL
      CREATE TRIGGER update_state_changed_at BEFORE UPDATE ON public.admins FOR EACH ROW WHEN (((old.state)::text <> (new.state)::text)) EXECUTE PROCEDURE update_state_changed_at()
  SQL
  create_trigger :update_state_changed_at, sql_definition: <<-SQL
      CREATE TRIGGER update_state_changed_at BEFORE UPDATE ON public.users FOR EACH ROW WHEN (((old.state)::text <> (new.state)::text)) EXECUTE PROCEDURE update_state_changed_at()
  SQL
end
```
It works same with sql schema (not rb). Can you give me more information about required changes?

teoljungberg:
If you create a function of name A with no parameters and then another function of name A with parameters. Can you dump the schema in rails have both functions present in the schema?

QNester:
Thanks, i've got it. I will research and I'll fix it if required.

QNester:
I have only one problem with `create_trigger` in case when multiple `create_trigger` using same SQL trigger but different params: migration can not be applied because 
```trigger "update_state_changed_at" for realtion "users" already exists```.

But trigger name (like `update_state_changed_at`) is defining in SQL definition. I think that good way to change trigger name in sql definition is not exists. Implicitly changing trigger name  can be so unexpectable for users and may cause many problems. 

I think that user should resolve this issue himself: if he want to reuse parametrized sql definitions with different params he must pay attention on trigger name in sql definition.  For example users can do somthing like:
```sql
CREATE TRIGGER update_state_changed_at_on_%{status_column}
    BEFORE UPDATE ON %{table}
    FOR EACH ROW
    WHEN (OLD.%{status_column} <> NEW.%{status_column})
EXECUTE PROCEDURE update_state_changed_at();
```
I think that best desicion is describe this behavior in readme.

And second problem - problem with `drop_trigger`. `drop_trigger` accepts argument `name` as name from sql_definiton. We have no mechanism to interpolate this name for dropping. And (again) I think that bast desicion - describe this beahavior in readme.  "If you need drop parametrized trigger you should pass interpolated trigger name to `drop_tirgger`". 

For example:
```sql
CREATE TRIGGER update_state_changed_at_on_%{status_column}
    BEFORE UPDATE ON %{table}
    FOR EACH ROW
    WHEN (OLD.%{status_column} <> NEW.%{status_column})
EXECUTE PROCEDURE update_state_changed_at();
```

```ruby
class CreateTriggerUpdateStateChangedAt < ActiveRecord::Migration[5.0]
  def up
    [User, Admin].each do |model|
      create_trigger :update_state_changed_at, on: model.table_name, params: { table: model.table_name, status_column: :state }
      create_trigger :update_state_changed_at, on: model.table_name, params: { table: model.table_name, status_column: :another_state }
    end
  end

  def down
      [User, Admin].each do |model|
           create_trigger :update_state_changed_at_on_state, on: model.table_name
           create_trigger :update_state_changed_at_on_another_state, on: model.table_name
      end
  end
end
```

What do you think about it?

teoljungberg:
What use cases are there for triggers with parameters? And do we need it now?

I wonder, can we complete functions first by having fx support creating functions with parameters, as well as importing existing schemas that has them. 
That's the major feature request I've gotten, such as #8

QNester:
I think that you right - maybe we don't need full support trigger parameters. But I need feature to pass table name to sql definition (this is reason why I thought use `on` option for interpolation).

Can I change this PR only for interpolate `table_name`?

teoljungberg:
Let's go with `on` as that's how the sql looks.

QNester:
I have just update it. Check it please and give me feedback.

QNester:
ping

teoljungberg:
I'll get to this in due time.

QNester:
it's bugged
I'll reopen it after fix

PR #58, opened 2020-04-27
Fixes for Rails edge

sobrinho:
Follow the same technique of thoughtbot/scenic, otherwise, we include our modules too soon and rails fail to boot.


sobrinho:
@teoljungberg one more for you.

teoljungberg:
I'm into it! Thanks

PR #59, opened 2020-08-31
Use current ActiveRecord version in migrations

dnamsons:


teoljungberg:
Excellent, many thanks @dnamsons.

PR #60, opened 2020-10-09
Support --no-migration generator flag

bf4:
Based on Rails [Model generator](https://github.com/rails/rails/blob/6-0-stable/activerecord/lib/rails/generators/active_record/model/model_generator.rb#L12)

I couldn't get the test suite to run locally

My use case is mostly that I'm just starting to use fx in my project and most of the time I want to generator a v1 before I use fx to change it, and I keep having to delete the migrations generated


teoljungberg:
Looks legit, many thanks @bf4!

PR #62, opened 2020-10-19
Fix: Support --no-migration generator flag

bf4:
- Fix impl and tests in https://github.com/teoljungberg/fx/pull/60

I figured out how to run the tests (on reflection, it's not hard, so don't know what my issue was before, maybe just appraisal

basically I (on Ruby 2.6.2)

```sh
export BUNDLE_GEMFILE=gemfiles/rails60.gemfile
gem install bundler --conservative
bundle check || bundle install
bundle exec rake dummy:db:drop
bundle exec rake dummy:db:create
bundle exec rake
```

### DIFF
--- a/spec/generators/fx/trigger/trigger_generator_spec.rb
+++ b/spec/generators/fx/trigger/trigger_generator_spec.rb
@@ -19,7 +19,7 @@ describe Fx::Generators::TriggerGenerator, :generator do
       migration = file("db/migrate/create_trigger_test.rb")
       trigger_definition = file("db/triggers/test_v01.sql")
 
-      run_generator ["test", "--no-migration"]
+      run_generator ["test", {"table_name" => "some_table"}, "--no-migration"]
 
       expect(trigger_definition).to exist
       expect(migration_file(migration)).not_to exist
--- a/lib/generators/fx/function/function_generator.rb
+++ b/lib/generators/fx/function/function_generator.rb
@@ -111,8 +111,9 @@ module Fx
         !migration
       end
 
+      # True unless explicitly false
       def migration
-        options[:migration]
+        options[:migration] != false
       end
     end
   end
--- a/lib/generators/fx/trigger/trigger_generator.rb
+++ b/lib/generators/fx/trigger/trigger_generator.rb
@@ -121,8 +121,9 @@ module Fx
         !migration
       end
 
+      # True unless explicitly false
       def migration
-        options[:migration]
+        options[:migration] != false
       end
     end
   end


teoljungberg:
The setup instructions are all documented in the CONTRIBUTING docs (which mentions the `bin/setup` script that does all of the steps above).

teoljungberg:
To avoid any further mishaps, I'm going to clone down the branch and play around with it to ensure that everything works as expected.

I'll get back to you with my findings, if any, @bf4.

bf4:
Sorry about that. In between when I submitted it and when you merged it I found the bug.

spec/generators/fx/trigger/trigger_generator_spec.rb
```diff
@@ -19,7 +19,7 @@
       migration = file("db/migrate/create_trigger_test.rb")
       trigger_definition = file("db/triggers/test_v01.sql")
 
-      run_generator ["test", "--no-migration"]
+      run_generator ["test", {"table_name" => "some_table"}, "--no-migration"]
 
       expect(trigger_definition).to exist
```
teoljungberg, line 24: If we do `--no-migration` - should the trigger exist then? To me, it feels like it shouldn't. We did ask it not to migrate... Maybe that flag is wrong? Or feels wrong?

bf4: The feature I was intending to add is **optionally* not (yet) generate a migration when generating the trigger.

The use case for me that is probably the best argument is that we just introduced fx to our app and when I want to update a function or trigger, I 1) first use fx to create the v1 function or trigger file, 2) delete the generated migration 3) use fx again to create the v2 function or trigger file 4) run the migration

The secondary use-case is when I am updating a trigger and the related function in one migration instead of two (or more), but that's my preference, and not as strong as an argument.

bf4: fwiw, I've bundled from this branch and it works as intended

```
📝 $ bin/rails g fx:function some_function --no-migration
      create  db/functions/some_function_v01.sql
ruby-2.6.2 ~/code
📝 $ bin/rails g fx:function some_other_function
      create  db/functions/some_other_function_v01.sql
      create  db/migrate/20201021032755_create_function_some_other_function.rb
```

<Details>

<Summary>
Gemfile
</Summary>

```ruby
# function migration helper: see https://github.com/teoljungberg/fx
# Bundle from github until merged https://github.com/teoljungberg/fx/pull/62
git 'https://github.com/bf4/fx', branch: 'add_skip_migration_generator_boolean' do
  gem "fx"
end
```

Gemfile.lock

```ruby
GIT
  remote: https://github.com/bf4/fx
  revision: 8b5b3705e1e3904a8fa3ffca99eabe157997134e
  branch: add_skip_migration_generator_boolean
  specs:
    fx (0.6.0)
      activerecord (>= 4.0.0)
      railties (>= 4.0.0)
```

<Details>

teoljungberg: Ah, sure - that makes sense to me. Thanks for explaining

bf4: Great!  I'll share back some other stuff I've been working on

PR #63, opened 2020-10-22
Expose migration methods for getting the sql definition

bf4:
Useful when you want to interact with the sql in various ways.

<Details>

<Summary>
For example, I use `strong_migrations`, so I made a mixin for all my
migrations to get the function sql.
</Summary>

```ruby
  # Could be simplified after https://github.com/teoljungberg/fx/pull/52 is merged
  def db_update_function(function_name, version:, revert_to_version:, replace: true)
    if replace
      reversible do |dir|
        dir.up do
          safely_execute function_sql(function_name, version: version)
        end
        dir.down do
          safely_execute function_sql(function_name, version: revert_to_version)
        end
      end
    else
      safety_assured do
        update_function function_name version: version, revert_to_version: revert_to_version
      end
    end
  end
```

</Details>

I expect you may want a different name for the methods than
`function_sql` and `trigger_sql`, but am using that as a discussion
point.  Notably, even though the trigger generator requires a table
name, it doesn't appear to be used.  (I've noticed an old issue regarding
this.)

Tests pass. `bin/setup; bin/appraisal rake` on CRuby 2.6.2.

### DIFF
--- a/lib/fx/statements/function.rb
+++ b/lib/fx/statements/function.rb
@@ -37,7 +37,7 @@ module Fx
           )
         end
         sql_definition = sql_definition.strip_heredoc if sql_definition
-        sql_definition ||= Fx::Definition.new(name: name, version: version).to_sql
+        sql_definition ||= function_sql(name, version: version)
 
         Fx.database.create_function(sql_definition)
       end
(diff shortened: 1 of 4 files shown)


bf4:
Anything you need from me?

teoljungberg:
I can see this being useful for strong_migrations, but I'd we can already do this by:

```ruby
Fx::Adapters::Postgres.new(connection).functions.find { |function| funcition.name == :test }.definition
```

or similar. But that should be enough until we have more use-cases for this.

PR #68, opened 2021-03-31
Dev 401/ruby 3 solo upgrade

nwmartin:
DON'T ACTUALLY MERGE THIS, IT IS GETTING A TAG

Basically what we've done here is updated the original fx repo with changes to actually make it work with ruby 3 (they have busted function calls) and then reapplied the aggregate changes over the top of it.


nwmartin:
Wrong base branch. Sorry!

PR #69, opened 2021-05-28
Add support for views and materialized views

comogo:
Hello folks! 

I using this gem in my current job and we have a lots of views and materialized views here. Rails does not support views when dumping the definitions of database into `schema.rb`, so I decided to fork this repository to add support for views and materialized views.

Materialized views can have indexes so when dumping data from materialized views it dumps that indexes definitions too.

I don't know if this feature makes sense for the main goal of the gem, but I'm just sharing the feature because it can be useful for someone else.

This is just the initial idea but it is already working and we are currently using it here.


teoljungberg:
I'd suggest using Scenic for all your database view needs.

https://github.com/scenic-views/scenic

comogo:
I didn't know that gem, I will try it out, thank you.

Issue #65, opened 2021-02-15
Ruby 3 support


incubus:
Hi!

  drop_trigger does not work with Ruby 3. Unfortunately, I couldn't make a quick fix for this.
  
```$ rake db:rollback
== 20210214133450 CreateTriggerUpdateOrderSubtotal: reverting =================
-- drop_trigger(:update_order_subtotal, {:on=>:orders_items})
rake aborted!
StandardError: An error has occurred, this and all later migrations canceled:

wrong number of arguments (given 2, expected 1; required keyword: on)


Caused by:
ArgumentError: wrong number of arguments (given 2, expected 1; required keyword: on)

Tasks: TOP => db:rollback
(See full trace by running task with --trace)
```

Thanks!

teoljungberg:
Yeah that's definitely a problem, and one I would like to see fixed.

teoljungberg:
As you expressed wanting to help @pjungwir - is this something you can help get over the finish line?

pjungwir:
I'd be happy to work on this! I want to add Ruby 3 to our Travis/Appraisal files too.

teoljungberg:
See https://github.com/teoljungberg/fx/pull/70 for starters @pjungwir, and if there's anything we ought to re-use.

pjungwir:
This should be fixed by https://github.com/teoljungberg/fx/pull/76.

teoljungberg:
Done as of #76 .

PR #70, opened 2021-07-06
Support for ruby 3

kadru:
Hi

This is my intent to fix ArgumentError: wrong number of arguments (given 2, expected 1; required keyword: on) for ruby 3 reported in  #65,  also fixes the failing tests for rails 4.2.

In my local two tests fails: `spec/fx/adapters/postgres/triggers_spec.rb:7` and `spec/fx/schema_dumper/trigger_spec.rb:4`, but I hope is because postgres version, I have postgres 12 installed.

### DIFF
--- a/lib/fx/statements/function.rb
+++ b/lib/fx/statements/function.rb
@@ -29,7 +29,10 @@ module Fx
       #     $$ LANGUAGE plpgsql;
       #   SQL
       #
-      def create_function(name, version: 1, sql_definition: nil)
+      def create_function(name, options = {})
+        version = options.fetch(:version, 1)
+        sql_definition = options[:sql_definition]
+
         if version.nil? && sql_definition.nil?
           raise(
             ArgumentError,
@@ -53,7 +56,8 @@ module Fx
       # @example Drop a function, rolling back to version 2 on rollback
       #   drop_function(:uppercase_users_name, revert_to_version: 2)
       #
-      def drop_function(name, revert_to_version: nil)
+      def drop_function(name, options = {})
+        revert_to_version = options[:revert_to_version]
         Fx.database.drop_function(name)
       end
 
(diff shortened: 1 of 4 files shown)


lib/fx/statements/function.rb
```diff
@@ -29,7 +29,10 @@ module Function
       #     $$ LANGUAGE plpgsql;
       #   SQL
       #
-      def create_function(name, version: 1, sql_definition: nil)
```
teoljungberg, line 32: I wonder if it's enough to use https://github.com/ruby/ruby2_keywords instead?

lib/fx/statements/trigger.rb
```diff
@@ -98,7 +103,12 @@ def drop_trigger(name, on:, revert_to_version: nil)
       #         EXECUTE PROCEDURE uppercase_users_name();
       #    SQL
       #
-      def update_trigger(name, version: nil, on: nil, sql_definition: nil, revert_to_version: nil)
+      # def update_trigger(name, version: nil, on: nil, sql_definition: nil, revert_to_version: nil)
```
teoljungberg, line 106: This seems accidental :)

lib/fx/statements/function.rb
kadru: Seems that doesn't  not fixes the errors, also print warnings: 
```
/home/oscar/workspace/fx/lib/fx/statements/function.rb:32: warning: Skipping set of ruby2_keywords flag for create_function (method accepts keywords or method does not accept argument splat)
```
I have used it this way:
```ruby
ruby2_keywords def create_function(name, version: 1, sql_definition: nil)
# rest of code...
```

lib/fx/statements/trigger.rb
kadru: Sorry I wil remove that.

pjungwir:
Thank you for this PR! I did pretty much the same thing in #76 except it's based on our latest changes and includes updates to the Travis & Appraisal files.

teoljungberg:
Closing in favor of #76 .

PR #71, opened 2021-07-08
Preserve backslashes when dumping the schema

pjungwir:
This is an alternative to https://github.com/teoljungberg/fx/pull/51 that fixes the same problem, but in a way that is more obviously-correct.

If a Postgres function contains backslashes (common in regexes e.g. \s
or \1), then we write those backslashes into a heredoc in schema.rb
without escaping them. So if you load your database from schema.rb, ruby
interprets the backslashes instead of preserving them for Postgres.
Rather than trying to escape every possible thing, this commit uses a
single-quote heredoc (<<-'SQL') which treats all characters literally.

### DIFF
--- a/lib/fx/function.rb
+++ b/lib/fx/function.rb
@@ -17,7 +17,7 @@ module Fx
 
     def to_schema
       <<-SCHEMA.indent(2)
-create_function :#{name}, sql_definition: <<-\SQL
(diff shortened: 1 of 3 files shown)


teoljungberg:
Ah this is great. Very clean.

PR #72, opened 2021-07-10
Update tests for new versions of Ruby & Rails

pjungwir:
This fixes various things so we can successfully run tests on Ruby 2.4, 2.5, 2.6, and 2.7. It also adds testing against Rails 6.1.

I broke things down into separate commits, but I can also just roll it all into one if you like.

On the other hand I thought a single PR was best since we aren't going to get a green from Appraisal/Travis until everything is combined.


teoljungberg:
This is wonderful! Thank you so much.

Issue #81, opened 2022-01-12
Issue for when the table_name_prefix or table_name_suffix are used


cavi21:
Hi! not sure if I'm missing something (maybe a config or similar) but when a project uses `table_name_prefix` (or `table_name_suffix`) on the `application.rb`, like:
```ruby
module ProjectName
  class Application < Rails::Application
    # ...
    config.active_record.table_name_prefix = 'project-prefix-'
    # ...
  end
end
```

Seems that the generated `functions` or `triggers` should be aware of, in such the seems to need to generate the files like:
```
/functions/project-prefix-logidze_capture_exception_v01.sql
...
/triggers/project-prefix-logidze_on_some_table_v01.sql
``` 
And as metioned in the [Issue #207 on the logidze gem](https://github.com/palkan/logidze/issues/207) this issue is related to how `fx` implemented the `create_function` method. It is though the `method_missing` defined in [activerecord-6.1.4.4/lib/active_record/migration.rb:915](https://github.com/rails/rails/blob/e5d046737e7dca41cfd3db5695d84f8709db1faf/activerecord/lib/active_record/migration.rb#L917) and so in there the method `proper_table_name( )` is called, and in such when you have in the migration:
```ruby
      #...
      dir.up do
        create_function :logidze_capture_exception, version: 1
      end
      #...
```
When the migration is run, the file that looks for is `functions/project-prefix-logidze_capture_exception_v01.sql` and the same happes with the triggers so if in the migration there is:
```ruby
      #...
      dir.up do
        create_trigger :logidze_on_some_table, on: :some_table
      end
      #...
```
The file that looks is `triggers/project-prefix-logidze_on_some_table_v01.sql` and notice that the second argument is fine because the same `proper_table_name( )` is called on them so the actual table name is referenced, in this case `project-prefix-some_table`

Not sure if this is something that you have in mind or do you want to address it. If so, what approach do you prefer?
Thanks for all the work you put into this gem!

teoljungberg:
I have never seen this usage before, it's an interesting one. If there are active-record helpers for this - we should totally use them. I'd entertain a PR that fixes this.

teoljungberg:
Feel free to take a stab at implementing this in a PR, that would be grand!

cavi21:
perfect @teoljungberg I'll look into it this weekend!

cavi21:
[...]

PR #82, opened 2022-01-15
Add support to the use of table_name_prefix or table_name_suffix

cavi21:
Hi @teoljungberg I'm opening this PR in order to address issue #81.

I add a new method on the migration stack in order to prevent the `method_missing(method, *arguments, &block)` on the **activerecord** `/lib/active_record/migration.rb` being called. Also tweaks the specs to handle the `full_table_name` when needed. I'm not sure how do you want to address the specs as I added the handling for environment variables in order to test everything when the `table_name_prefix` and/or `table_name_suffix` are used. So all the specs can be run like
```bash
bin/appraisal rake
TABLE_NAME_PREFIX='prefix-' bin/appraisal rake
TABLE_NAME_SUFFIX='-suffix' bin/appraisal rake
```

But not sure if this is what you had in mind? Anything let me know

### DIFF
--- a/lib/fx/migration.rb
+++ b/lib/fx/migration.rb
@@ -1,0 +1,31 @@
+require "rails"
+
+module Fx
+  # @api private
+  module Migration
+
+    def fx_migration_action(*arguments, &block)
+      _method = __callee__
+      arg_list = arguments.map(&:inspect) * ", "
+
+      say_with_time "#{_method}(#{arg_list})" do
+        unless connection.respond_to? :revert
+          if !arguments.empty? && [:drop_trigger, :update_trigger].include?(_method)
+            if arguments.second.is_a?(Hash) && arguments.second.key?(:on)
+              on_table = arguments[1][:on]
+              arguments[1] = arguments.second.merge(on: proper_table_name(on_table, table_name_options))
+            end
+          end
+        end
+        return super unless connection.respond_to?(_method)
+        connection.send(_method, *arguments, &block)
+      end
+    end
+    alias_method :create_function, :fx_migration_action
+    alias_method :drop_function, :fx_migration_action
+    alias_method :update_function, :fx_migration_action
+    alias_method :create_trigger, :fx_migration_action
+    alias_method :drop_trigger, :fx_migration_action
+    alias_method :update_trigger, :fx_migration_action
+  end
+end
--- a/lib/fx.rb
+++ b/lib/fx.rb
@@ -4,6 +4,7 @@ require "fx/command_recorder"
 require "fx/configuration"
 require "fx/definition"
 require "fx/function"
+require "fx/migration"
 require "fx/statements"
 require "fx/schema_dumper"
 require "fx/trigger"
@@ -30,6 +31,11 @@ module Fx
     ActiveRecord::ConnectionAdapters::AbstractAdapter.send(
       :include,
       Fx::Statements,
+    )
+
+    ActiveRecord::Migration.send(
+      :include,
+      Fx::Migration,
     )
   end
 
--- a/lib/fx/adapters/postgres.rb
+++ b/lib/fx/adapters/postgres.rb
@@ -142,7 +142,7 @@ module Fx
       #
       # @return [void]
       def drop_trigger(name, on:)
-        execute "DROP TRIGGER #{name} ON #{on};"
+        execute "DROP TRIGGER #{name} ON \"#{on}\";"
       end
 
       private
--- a/spec/acceptance/user_manages_triggers_spec.rb
+++ b/spec/acceptance/user_manages_triggers_spec.rb
@@ -1,6 +1,13 @@
 require "acceptance_helper"
 
 describe "User manages triggers" do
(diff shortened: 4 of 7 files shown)


lib/fx/migration.rb
```diff
@@ -0,0 +1,31 @@
...
+        unless connection.respond_to? :revert
+          if !arguments.empty? && [:drop_trigger, :update_trigger].include?(_method)
+            if arguments.second.is_a?(Hash) && arguments.second.key?(:on)
+              on_table = arguments[1][:on]
+              arguments[1] = arguments.second.merge(on: proper_table_name(on_table, table_name_options))
+            end
+          end
+        end
+        return super unless connection.respond_to?(_method)
+        connection.send(_method, *arguments, &block)
```
cavi21, line 21: it's possible to remove the `&block` references in this action because `fx` is not using it. I was trying to be a similar implementation of the `method_missing` on activerecord, but anything let me know!

lib/fx/migration.rb
```diff
@@ -0,0 +1,31 @@
+require "rails"
+
+module Fx
+  # @api private
+  module Migration
+
+    def fx_migration_action(*arguments, &block)
```
teoljungberg, line 7: I'm not excited about the complexity this code brings, I wanted something cleaner where we leveraged the `ActiveRecord::Base.table_name_prefix` to coerce:

```
create_function :test_function
```

gjven that the table name prefix is `ActiveRecord::Base.table_name_prefix = "abc_"`, I'd expect that we named the function `abc_test_function`. How we do that, I'm less sure about. But I don't see this being the way we go about implementing this.

The same goes for the other `*_function` and `*_trigger` migration helpers.

This implementation feels too convoluted for something I'd like to maintain going forward.

teoljungberg (changes requested):

lib/fx/migration.rb
cavi21: Hi @teoljungberg thanks for looking into it! I went with this approach because I felt that it's more similar to the way Activerecord deals with the migrations, and schema in which they apply the `table_name_prefix` (and/or `table_name_suffix`) when the migration is run and not on the generate or dump. 

But if you rather keep using the `method_missing` way, as you said we need to rename the generated **functions** and **triggers**. I'm not sure if it's going to feel so consistent with other parts of active record but we can explore it for sure. 

Make sense to work with this other approach on another branch and open another PR, or what do you have in mind?

teoljungberg: Maybe it is the case that this fits best into how activerecord handles migrations, as you have done here, maybe it doesn't. I don't know enough to make a judgement call yet. However, the code as submitted now is not stable enough for something I'd maintain, it's more procedural than OO.

If we back up a bit, what's the use-case of using the prefix-and-suffixes? What do we want to accomplish here? If I create a function named `testing1_testing2` and have the prefix set to `abc_` and suffix to `_def`, should the function be created as `abc_testing1_testing2_def` in the database?

From where I'd tackle this, I'd see how it fits into `create_function` (to take one of the two code paths - functions vs triggers)

https://github.com/teoljungberg/fx/blob/946cdccbd12333deb8f4566c9852b49c0231a618/lib/fx/statements/function.rb#L7-L46

And they all point to Definition, so maybe that's a good place to inline this?

But to go back up to the original point, I don't understand _why_ we should support it, what the use cases are, and how they are expected to behave. I'd like to see those things adressed in the PR description before we move forward.

cavi21: > Maybe it is the case that this fits best into how activerecord handles migrations, as you have done here, maybe it doesn't. I don't know enough to make a judgement call yet. However, the code as submitted now is not stable enough for something I'd maintain, it's more procedural than OO.

I understand and make sense to don't focus on how activerecord is dealing with it and focus on the approach that `fx` has.

> If we back up a bit, what's the use-case of using the prefix-and-suffixes? 

One use case to set the `config.active_record.table_name_prefix = ...` is for when different services share the same database, so if two services for instance has the table `users` they don't collide each other, because each has their own `prefix` (or `suffix`)

> What do we want to accomplish here?

That's a good question! so in the case where the `table_name_prefix` the generation of the files will not fail, but when you run de migration does. Because they could not found the  `prefix-...` **function** or **trigger**. So thinking about it, that's why I focused on the migration run and not the generator part.

> (3) If I create a function named testing1_testing2 and have the prefix set to abc_ and suffix to _def, should the function be created as abc_testing1_testing2_def in the database? 

Another good question that I'm not sure what makes more sense.

Overall I understand that this adds complexity and weirdness to the hole gem, and maybe it's not something that should be added to it.  So feel free to close this PR if you think the same, or if you rather add support to this issue, following your suggestion, make sense to define the strategy or how to proceed. As you said could be at the `Definition` part or somewhere like it, and also the (3) needs to be anwsered, right?

Open to whatever you think is best for the gem!

cavi21:
Hi @teoljungberg I will close this PR based on our last messages seems that's reasonable. The issue #82 is still open if you think that make sense to have the conversation on how is the best way to approach this use case or if even make sense at all for the gem.

Thanks for all the feedback and time put into the gem!!

PR #88, opened 2022-05-13
Allow exclusion of some triggers/functions from schema

ScottLangridge:
In some cases it is necessary to prevent some functions or triggers from being included in schema.rb. 

An example of this is my use case, where I am using Timescale DB, which adds non-user defined triggers which bloat the schema when dumped. Additionally, my database is multitenant, meaning that the trigger is added once for every single tenant.

The addition of a configurable condition to control which functions and triggers make into the schema removes this issue.

### DIFF
--- a/lib/fx/configuration.rb
+++ b/lib/fx/configuration.rb
@@ -40,9 +40,25 @@ module Fx
     # @return Boolean
     attr_accessor :dump_functions_at_beginning_of_schema
 
+    # A lambda returning a boolean to indicate whether or not a given trigger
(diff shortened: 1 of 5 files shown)


ScottLangridge:
^ Updated PR description for clairty.

teoljungberg:
I wonder, can we instead update the filtering mechanism we use to get the functions from the database with? And only getting those owned by the current user or something similar?

That seems like a better solution than having an inclusion / exclusion mechanism to me.

teoljungberg:
I have not heard back, so I'm closing this.

PR #94, opened 2022-07-19
Add dumping at beginning/end of schema

zedevs:
We needed a more flexible way to order the schema because we had different types of dependencies (veiws depend on functions, functions depend on views and so on), and the simple approach of dump on top or at the bottom just didn't work for us. This is what we came up with we've been using it for a couple of months now. 

Before committing to writing specs and cleaning it up, what do you folks think about it? Is this a feature you would like to add to fx?

### DIFF
--- a/lib/fx/configuration.rb
+++ b/lib/fx/configuration.rb
@@ -39,10 +39,14 @@ module Fx
     # Defaults to false
     # @return Boolean
     attr_accessor :dump_functions_at_beginning_of_schema
+    attr_accessor :functions_to_dump_at_beginning_of_schema
+    attr_accessor :functions_to_dump_at_end_of_schema
 
     def initialize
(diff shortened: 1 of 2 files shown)


teoljungberg:
This seems a little bit complex to me as a configuration option. And, I have not seen a need for this in the wild, yet.

Are there any real life examples you can share that adds weight to this argument?

teoljungberg:
I'm also curious to hear why not having all functions in the beginning of the schema is not enough for this use-case.

teoljungberg:
I converted this PR to a draft as it is not mergable without tests, and doesn't need to be before the idea is discussed more.

It's only for my own sanity so I know where to focus my energy.

teoljungberg:
Closing for now, I'd be happy to discuss this further when more information is on the table.

PR #80, opened 2021-12-01
Add a link to F(x) SqlServer Adapter in the README

tarellel:

PR #95, opened 2022-07-30
Fix forward references when loading schema

pjungwir:
If your functions depend on each other, you may wind up with forward
references in your `schema.rb` file (even though we order them by oid,
e.g. if an older function was refactored to call new functions).
Postgres has the same problem in `pg_dump`, and they solve it by
temporarily disabling a GUC called `check_function_bodies`. This commit
adds calls to `schema.rb` to do the same thing.

I haven't added any new tests yet (though old ones all pass), and I haven't thought about how this breaks non-Postgres adapters (if there are any out there). So this is just a quick fix I can use personally, but if you would like a more polished version please let me know if you agree with the general approach here.

### DIFF
--- a/lib/fx/schema_dumper/function.rb
+++ b/lib/fx/schema_dumper/function.rb
@@ -23,9 +23,11 @@ module Fx
       end
 
       def functions(stream)
+        stream.puts('  execute "SET check_function_bodies=off"')
         dumpable_functions_in_database.each do |function|
           stream.puts(function.to_schema)
         end
+        stream.puts('  execute "SET check_function_bodies=on"')
       end
 
       private


lib/fx/schema_dumper/function.rb
```diff
@@ -23,9 +23,11 @@ def empty_line(stream)
       end
 
       def functions(stream)
+        stream.puts('  execute "SET check_function_bodies=off"')
```
teoljungberg, line 26: If we are going down this route, this should be performed on the adapter level to have a `{pre,post}_function_dump`.

I have yet to think about this, or seen it. So there is some reading up for me to do.

teoljungberg:
As we do have other adapters after going through old issues:

https://github.com/teoljungberg/fx/pull/80

Meaning, we need to have this be database-specific if we are going to go forward.

I'll close this PR until that is handled, if we are going to handle it.

PR #98, opened 2022-12-24
Add ERB Support for SQL definitions

gmhawash:
Allow for ERB interpolation in SQL definition files.  This is helpful when SQL definition relies on some ruby constants or methods.

### DIFF
--- a/lib/fx/definition.rb
+++ b/lib/fx/definition.rb
@@ -8,7 +8,7 @@ module Fx
     end
 
     def to_sql
-      File.read(find_file || full_path).tap do |content|
+      ERB.new(File.read(find_file || full_path)).result.tap do |content|
         if content.empty?
           raise "Define #{@type} in #{path} before migrating."
         end
--- a/lib/fx/definition.rb
+++ b/lib/fx/definition.rb
@@ -3,7 +3,7 @@ module Fx
   class Definition
     def initialize(name:, version:, type: "function")
       @name = name
-      @version = version.to_i
+      @version = version
       @type = type
     end
 
@@ -24,7 +24,8 @@ module Fx
     end
 
     def version
-      @version.to_s.rjust(2, "0")
+      @version = latest_version if @version == :latest
+      @version.to_i.to_s.rjust(2, "0")
     end
 
     private
@@ -41,6 +42,11 @@ module Fx
 
     def migration_paths
       Rails.application.config.paths["db/migrate"].expanded
+    end
+
+    def latest_version
+      Dir.glob(Rails.root.join("db", @type.pluralize, "#{@name}*.sql")).max =~ /#{@name}_v(\d*).sql/i
+      $1
     end
   end
 end
--- a/lib/fx/version.rb
+++ b/lib/fx/version.rb
@@ -1,4 +1,4 @@
 module Fx
   # @api private
-  VERSION = "0.7.0"
+  VERSION = "1.7.1"
 end
--- a/lib/fx/version.rb
+++ b/lib/fx/version.rb
@@ -1,4 +1,4 @@
 module Fx
   # @api private
-  VERSION = "1.7.1"
+  VERSION = "1.7.2"
 end
--- a/lib/fx/version.rb
+++ b/lib/fx/version.rb
@@ -1,4 +1,4 @@
 module Fx
   # @api private
-  VERSION = "1.7.2"
+  VERSION = "0.7.0"
 end
--- a/spec/fx/definition_spec.rb
+++ b/spec/fx/definition_spec.rb
@@ -25,7 +25,7 @@ describe Fx::Definition do
 
         expect { definition.to_sql }.to raise_error(
           RuntimeError,
-          %r(Define function in db/functions/test_v01.sql before migrating),
+          %r{Define function in db/functions/test_v01.sql before migrating}
         )
       end
 
@@ -99,7 +99,7 @@ describe Fx::Definition do
         definition = Fx::Definition.new(
           name: "test",
           version: 1,
-          type: "trigger",
+          type: "trigger"
         )
 
         expect(definition.to_sql).to eq sql_definition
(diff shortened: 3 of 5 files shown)


teoljungberg:
I have never seen this in real life practice, can you cite a project or two that uses this?

gmhawash:
Our application uses a private gem that exposes a set of transaction codes.  Every once in a while we update the gem (based on customer requirements) and the transaction codes are used in SQL functions, views and queries.  So for us, rather than listing the transactions codes all over the place; we do this inside these SQL entities:

```
SELECT ....
WHERE transactions.code  IN (<%= TransactionGem.payment_codes.join(',') %>) OR transactions.code in (<%= TransactionGem.jouranl_entry_codes.join(',') %>)
```

And when we update the transaction gem; we just rerun the migration to recreate the SQL functions and views and that updates them properly.

teoljungberg:
I see! Thanks for context.

I'm going through a maintenance of the project to cut out old and stale versions of ruby and rails. And will cut a release for that before I come back to this.

lib/fx/version.rb
```diff
@@ -1,4 +1,4 @@
 module Fx
   # @api private
-  VERSION = "0.7.0"
+  VERSION = "1.7.1"
```
teoljungberg, line 3: You don't need to change this, I'll do that once a release is ready.

```suggestion
  VERSION = "0.7.0"
```

Or to be on top of master

```suggestion
  VERSION = "0.8.0"
```

spec/fx/definition_spec.rb
```diff
@@ -29,6 +29,39 @@
...
+            RETURNS text AS $$
+            BEGIN
+                RETURN '<%= function %>';
+            END;
+            $$ LANGUAGE plpgsql;
+          <% end %>
+        EOS
+
+        erb_sql_definition = (<<-EOS).delete("\n").squeeze(' ').strip
+            CREATE OR REPLACE FUNCTION test()
```
teoljungberg, line 45: Swap this to 2 space indentation

spec/fx/definition_spec.rb
```diff
@@ -29,6 +29,39 @@
...
+            BEGIN
+                RETURN 'test';
+            END;
+            $$ LANGUAGE plpgsql;
+            CREATE OR REPLACE FUNCTION west()
+            RETURNS text AS $$
+            BEGIN
+                RETURN 'west';
+            END;
+            $$ LANGUAGE plpgsql;
```
teoljungberg, line 56:
```suggestion
          CREATE OR REPLACE FUNCTION test()
          RETURNS text AS $$
          BEGIN
              RETURN 'test';
          END;
          $$ LANGUAGE plpgsql;
          CREATE OR REPLACE FUNCTION west()
          RETURNS text AS $$
          BEGIN
              RETURN 'west';
          END;
          $$ LANGUAGE plpgsql;
```

gmhawash:
I pushed a commit to resolve these issues.

spec/fx/definition_spec.rb
```diff
@@ -8,7 +8,7 @@
           CREATE OR REPLACE FUNCTION test()
           RETURNS text AS $$
           BEGIN
-              RETURN 'test';
```
teoljungberg, line 11: Thanks for fixing these too 👏🏻

spec/fx/definition_spec.rb
```diff
@@ -29,13 +29,46 @@
...
+          BEGIN
+            RETURN 'west';
+          END;
+          $$ LANGUAGE plpgsql;
+        EOS
+        allow(File).to receive(:read).and_return(sql_definition)
+
+        definition = Fx::Definition.new(name: "test", version: 1)
+
+        expect(definition.to_sql.delete("\n").squeeze(' ').strip).to eq erb_sql_definition
```
teoljungberg, line 62: This expectation is more than a mouthful to read. Not a blocker for now, but I'll take a stab at reformatting it to make it leaner and meaner

teoljungberg (approved):

teoljungberg:
I ran CI @gmhawash - and there are some `standardrb` fixes to be made before we merge this.

teoljungberg (changes requested):
As mentioned above - CI is failing.

gmhawash:
When I run this locally everything passes.  What do I need to do to make it fail locally so I can fix these issues?

teoljungberg:
@gmhawash it’s because your local version of standard.rb is out of date with ci

spec/fx/definition_spec.rb
```diff
@@ -29,13 +29,46 @@
...
+          CREATE OR REPLACE FUNCTION west()
+          RETURNS text AS $$
+          BEGIN
+            RETURN 'west';
+          END;
+          $$ LANGUAGE plpgsql;
+        EOS
+        allow(File).to receive(:read).and_return(sql_definition)
+
+        definition = Fx::Definition.new(name: "test", version: 1)
```
teoljungberg, line 60:
```suggestion
        definition = Fx::Definition.function(name: "test", version: 1)
```

AS this was changed on master with #119

spec/fx/definition_spec.rb
```diff
@@ -29,13 +29,46 @@
...
+          <% %w(test west).each do |function| %>
+            CREATE OR REPLACE FUNCTION <%= function %>()
+            RETURNS text AS $$
+            BEGIN
+              RETURN '<%= function %>';
+            END;
+            $$ LANGUAGE plpgsql;
+          <% end %>
+        EOS
+
```
teoljungberg, line 43:
```suggestion
```

lib/fx/definition.rb
```diff
@@ -24,7 +24,8 @@ def path
     end
 
     def version
-      @version.to_s.rjust(2, "0")
+      @version = latest_version if @version == :latest
```
teoljungberg, line 27: This feels unrealted to this PR?

teoljungberg (changes requested):

teoljungberg:
Having thought about this more from a consistency perspective, this change would make the database migration non-reversible as the evaluated code with ERB can and will probably change.

I would be surprised if that happened to me, and therefore don't want to include that uncertainty.

More reasoning along the same lines from fx's sister project scenic:
https://github.com/scenic-views/scenic/issues/248

Issue #91, opened 2022-06-15
Drop EOL Ruby versions


teoljungberg:
Ruby's EOL is <= 2.6, we should drop the versions older than that, and ensure that 2.7 and above are supported.

PR #110, opened 2023-01-21
Bump Ruby to 2.7 and above

teoljungberg:


teoljungberg:
Resolves https://github.com/teoljungberg/fx/issues/91

Issue #40, opened 2019-08-21
Drop EOL'd Rails versions


teoljungberg:
As Rails 6 is shipped(#39) - we want to drop support for Rails 4.2 and ruby versions older than 2.2.2.

petergoldstein:
@teoljungberg What's required to do this?  And what's the Rails / Ruby support policy for this gem?

I'd like to add Rails 7 and Ruby 3.1, and that's a lot easier with a narrower support matrix.  Would you be comfortable dropping everything before Rails 5.2 and Ruby 2.6?

teoljungberg:
I have not written a formal policy, so we can rely on Rails' own: https://guides.rubyonrails.org/maintenance_policy.html

That means, we can drop everything before 5.2.x now.

Do you want to setup a PR for that?

teoljungberg:
1. I'd want to support all currently supported ruby versions. I'm fine with that. But we should drop Rails and Ruby versions in separate PRs. Ideally, releaseing them separately too.
2. Huh, that's odd. No I've only relied on Travis' open source version. But now that you mention it, it's odd. Did they cut off all open source plans? I don't see anything being mentioned of that on their website, but it could be. If travisci is no longer free, I'd love to switch to github actions.

This project sees little development as I've been happy about the state that it's in, so that could be why we haven't noticed the changes to TravisCI's setup. But I'm open for changing, and yes please if you want to whip up a PR - I'd love that.

teoljungberg:
Created #92 for the replacement of Travis, and #91 for the dropping of unsupported Ruby versions.

teoljungberg:
Resolves #

PR #111, opened 2023-01-21
Bump Rails to 6.0 and above

teoljungberg:


teoljungberg:
Resolves https://github.com/teoljungberg/fx/issues/40

PR #115, opened 2023-03-16
Fix generating function migrations for functions in custom schemas

GUI:
Currently things don't quite work right if you try to generate a function migration for a function inside a custom PostgreSQL schema. The underlying `create_function` (and other methods) work with a schema-specified name, however the migration file names and file naming assumptions get confused when a schema name is in the mix.

For example, if you currently run a command like this:

```
rails generate fx:function my_schema.my_function
```

You'll end up with `db/functions/my_schema.my_function_v01.sql` and `db/migrate/[timestamp]_create_function_my_schema.my_function.rb` files. The extra dot in the migration filename causes issues for Rails, as does the invalid class name it uses inside (`class CreateFunctionMySchema.myFunction < ActiveRecord::Migration[7.0]`).

This fixes the issue by replacing dots in the file name/table name with underscores, which allows for the normal Rails filename and class name conventions to work. But the original name, with the dot, is still retained and used inside the migration itself for the `create_function` method call, so the underlying SQL for creating and rolling back functions is correctly schema-specified.

This adds test coverage for this function generator handling schema-specified table names. It also adds test coverage for schema-specified table names being used with the trigger generator, but that all worked already since for triggers the table name wasn't present in the file name, so this wasn't an issue.

### DIFF
--- a/lib/fx/definition.rb
+++ b/lib/fx/definition.rb
@@ -41,7 +41,7 @@ module Fx
     private
 
     def filename
-      @_filename ||= "#{@name}_v#{version}.sql"
+      @_filename ||= "#{@name.to_s.tr(".", "_")}_v#{version}.sql"
     end
 
     def find_file
--- a/lib/generators/fx/function/function_generator.rb
+++ b/lib/generators/fx/function/function_generator.rb
@@ -54,6 +54,15 @@ module Fx
           @_version ||= previous_version.next
         end
 
+        alias_method :original_file_name, :file_name
+        def file_name
+          super.tr(".", "_")
+        end
+
+        def singular_name
+          original_file_name
+        end
+
         def migration_class_name
           if updating_existing_function?
             "UpdateFunction#{class_name}ToVersion#{version}"
--- a/spec/fx/definition_spec.rb
+++ b/spec/fx/definition_spec.rb
@@ -87,6 +87,12 @@ describe Fx::Definition do
 
         expect(definition.path).to eq "db/functions/test_v01.sql"
       end
+
+      it "uses underscores for schema-specified names" do
+        definition = Fx::Definition.new(name: "foo.test", version: 15)
+
+        expect(definition.path).to eq "db/functions/foo_test_v15.sql"
+      end
     end
 
     context "representing a trigger definition" do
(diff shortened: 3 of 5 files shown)


teoljungberg:
This seems sensible, and a good addition to normalize the input.

Reading through the diff I'm seeing a bit of duplication in how this is implemented, and this is to no fault of this PR - but how the design of f(x) was done initially.

I'd like to merge this, but will focus on getting a maintenance update out first with what's unreleased on master as well as #110 and #111.

Let me do that and I'll get back to this soon!

teoljungberg:
There are some CI issues to investigate @GUI relating to `rake standard:fix` - they are probably solved by a rebase on master or are an easy fix.

GUI:
@teoljungberg: Sorry for the delay, but I've rebased with `master`, and it looks like that's fixed the `standard` linting now.

Semi-related, I thought I had this passing the standard tests when I pushed this up (but maybe I missed something). I know at some point I saw a newer version of standard was released that I think changed some of the rules which broke things in CI. Would it be worth committing the `Gemfile.lock` to this repo so the CI environment is more reproducible (and the `standard` upgrades would then be more explicit)? I know originally, committing the `Gemfile.lock` was maybe discouraged for gem development, but I think that guidance has since changed: https://bundler.io/guides/faq.html#using-gemfiles-inside-gems

> **Q:** Should I commit my `Gemfile.lock` when writing a gem?
>
> **A:** Yes, you should commit it. The presence of a `Gemfile.lock` in a gem’s repository ensures that a fresh checkout of the repository uses the exact same set of dependencies every time. We believe this makes repositories more friendly towards new and existing contributors. Ideally, anyone should be able to clone the repo, run `bundle install`, and have passing tests. If you don’t check in your `Gemfile.lock`, new contributors can get different versions of your dependencies, and run into failing tests that they don’t know how to fix.
>
> **Q:** But I have read that gems should not check in the Gemfile.lock!
>
> **A:** The main advantage of not checking in your Gemfile.lock is that new checkouts (including CI) will immediately have failing tests if one of your dependencies changes in a breaking way. Instead of forcing every fresh checkout (and possible new contributor) to encounter broken builds, the Bundler team recommends either using a tool like Dependabot to automatically create a PR and run the test suite any time your dependencies release new versions. If you don’t want to use a dependency monitoring bot, we suggest creating an additional daily CI build that deletes the Gemfile.lock before running `bundle install`. That way you, and others monitoring your CI status, will be the first to know about any failures from dependency changes.

But anyway, just a thought. Thanks for reviewing  all this!

teoljungberg:
Regarding Gemfile.lock - I went down the same route sidekiq and zeitwerk has done it - where they have not checked the lockfile in.

Issue #114, opened 2023-02-20
Add support for create_operator


veganstraightedge:
Hi. Thanks for making `fx`! it helped me clean up the company's big rails app's migrations quite a bit.

# Currently

Sadly, we still have to use `db/structure.sql`.
I'd prefer if we could go back to using `db/schema.rb`.

The one last thing that stands between here and there is a custom operator.

Currently, we have this code in the same migration with all of our `create_function ...` calls:

```ruby
# TODO: figure out if/how we can express this so it ends up in schema.rb:
execute <<~SQL.squish
  CREATE OPERATOR ||+ (LEFTARG = TEXT,  RIGHTARG = TEXT, PROCEDURE = immutable_concat);
SQL
```

I'd rather have:

```ruby
create_operator :pipe_pipe_plus
```

***

This line of code:

```SQL
CREATE OPERATOR ||+ (LEFTARG = TEXT,  RIGHTARG = TEXT, PROCEDURE = immutable_concat);
```

is depended on by a few lines of code in migrations like this:

```ruby
t.index "(((((name)::text ||+ (location_line_1)::text) ||+ (location_city)::text) ||+ (location_region)::text)) gin_trgm_ops", name: "index_jobs_1_gin", using: :gin
```

I tried, just in case, doing a `create_operator :pipe_pipe_plus` along with all of our `create_function`s.  
This raises a `NoMethodError` for #create_operator, obviously.

I tried being sneaky and doing `create_function :pipe_pipe_plus`.
This didn't blow up, but then didn't end up in `db/schema.rb`.

I'm assuming that this is because `db/schema.rb` has no concept of `create_operator` like it has `create_function`, `create_trigger`, `create_view`, et al (?).

# Feature request 

Could you add a `create_operator` method that works just like `create_function`, which looks for a file named the same as the passed in symbol arg but in a `db/operators` folder instead of `db/functions`?

And then :hand wave: (I don't know what this would take), also get that to work in `db/schema.rb` please?

That would make me so happy.

teoljungberg:
I've never had to create custom operators, nor ever thought that I could.

I'm unsure if makes sense for F(x), but PostgreSQL mentions operators in the same breath as functions - so I guess it hold some merit to it.

I won't have the bandwidth to dig into this myself - but I'd entertain discussing this in a PR if someone opens it. I'm still not sure if this makes sense into F(x) proper or if it should be a gem on its own named `fx-operator` or similar.

veganstraightedge:
[...]

PR #116, opened 2023-03-17
Add support for SQL views and materialized views

GUI:
This resurrects @comogo work from https://github.com/teoljungberg/fx/pull/69, with just some minor updates to bring it up to date with `master`.

@teoljungberg, I know you originally closed #69 and recommended using Scenic instead. That makes sense, but I'm wondering if you'd be open to revisiting this due to some specific use-cases where fx's approach seems to make things easier. Namely, it revolves around similar needs brought up in https://github.com/scenic-views/scenic/issues/338 involving permissions on views, but it doesn't seem like Scenic has a desire to handle (which is fair).

Basically, for situations where you need to perform `GRANT` statements on materialized views, you need to execute those grants every time your materialized view is updated or replaced (since materialized views have to be dropped and then recreated, so it doesn't retain the original permissions). But here's how things compare between Scenic and fx and why fx with these changes feels like it works better for this use-case:

- In Scenic the SQL file is *only* the `SELECT` statement that defines the view. There cannot be any other SQL statements in the SQL file definition, so if you want to perform extra logic in the same transaction, like `GRANT` statements, then those  would need to be added to the Ruby migration file as `execute` statements.

  While that can work, the downside of this approach is that since these extra GRANT statements are in the migration file, that means this logic is not automatically copied when you add a new migration to update the view. Only the SQL file definition is copied between migration versions, so unless you're careful to track down the previous Ruby migration file and copy extra `execute` statements from there into your new migration file, it's easy for permissions to go missing when updating views.

  And I can understand why this would be a complicated feature to add to Scenic due to how the SQL file only defines the `SELECT` statement. So for similar reasons that supporting indices are tricky for materialized view (since Scenic needs to query the database for indices before and then re-create them afterwards), I can see why this would be more of a pain to handle in Scenic.
- With fx's approach and this work from @comogo, this all seems simpler to handle without jumping through extra hoops. Because fx will execute the SQL definition file fully (rather than making assumptions that it only includes the `SELECT` statement), then you simply need to include any `GRANT` or `CREATE INDEX` statements in the SQL definition file. Since any future update migrations will copy that full SQL file, then any custom permissions or indices will automatically be retained since they're simply part of the SQL definition file (so you don't need to have any custom logic in the Ruby migration file).

If this still feels like an edge-case you're not interested in supporting, then I understand, but it does seem like fx's approach does lend itself to handling views and materialized views in a nice way that isn't too complicated (since most of the details are being deferred to the SQL definition files, but we still can benefit from fx's versioning approach).

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,22 +4,22 @@
 [![Documentation Quality](http://inch-ci.org/github/teoljungberg/fx.svg?branch=master)](http://inch-ci.org/github/teoljungberg/fx)
 
 F(x) adds methods to `ActiveRecord::Migration` to create and manage database
-functions and triggers in Rails.
+functions, triggers and views in Rails.
 
-Using F(x), you can bring the power of SQL functions and triggers to your Rails
-application without having to switch your schema format to SQL. F(x) provides
-a convention for versioning functions and triggers that keeps your migration
-history consistent and reversible and avoids having to duplicate SQL strings
-across migrations. As an added bonus, you define the structure of your function
-in a SQL file, meaning you get full SQL syntax highlighting in the editor of
-your choice and can easily test your SQL in the database console during
-development.
+Using F(x), you can bring the power of SQL functions, triggers and views to your
+Rails application without having to switch your schema format to SQL. F(x)
+provides a convention for versioning functions, triggers and views that keeps
+your migration history consistent and reversible and avoids having to duplicate
+SQL strings across migrations. As an added bonus, you define the structure of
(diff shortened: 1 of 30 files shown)


teoljungberg:
I don't want to include view-support into f(x) as scenic exists to solve those exact needs, and I want to keep f(x) focused as a library to support functions and triggers. Not much more than that, maybe include operations (#114) as PostgreSQL mentions functions and operators in the same breath in their own documentation.

I'd suggest opening an issue to Scenic to discuss how the GRANT support can be included.

PR #124, opened 2023-04-24
Raise on warnings

teoljungberg:


teoljungberg:
This should catch stuff like #122 going onwards, hopefully

PR #127, opened 2023-05-29
Require Rails >= 6.1

teoljungberg:
Rails < 6.1 (and 6.0 specifically) hit EOL on 1/6/2023.


teoljungberg:
Rails < 6.1 (and 6.0) specifically hits EOL 1/6/2023.

PR #128, opened 2023-05-29
Require Ruby >= 3.0

teoljungberg:


teoljungberg:
Ruby < 3.0 (and 2.7) specifically hit EOL 31/3/2023.

PR #135, opened 2023-08-29
Implement check_function_bodies only for Postgres

cmer:


cmer:
Here's a different implementation of #134 . Specs don't pass (not sure why yet), but this is simpler and only for PG.

cmer:
Is this implementation more aligned with what you had in mind?

cmer:
Any chance this could be merged in? Thanks!

lib/fx/configuration.rb
```diff
@@ -40,9 +40,25 @@ class Configuration
...
+    # The default and recommended value is `false` to mimic pg_dump's behavior.
+    #
+    # Defaults to false
+    # @return Boolean, nil
+    attr_accessor :check_function_bodies
+
     def initialize
       @database = Fx::Adapters::Postgres.new
       @dump_functions_at_beginning_of_schema = false
+      @check_function_bodies = false
```
cmer, line 61: I suggest making this `true` by default.

teoljungberg:
I'm unsure I want to support this, because I have no context of the use-case or how it is supported in other databases (and therefore the other adapters).

cmer:
From the previous PR:

> Added a new configuration parameter check_function_bodies to allow controlling the database behavior when loading functions.
>
> This is necessary if the function created relies on the existence of a table that is not yet created, for example.
>
> Disabling check_function_bodies is pg_dump's default behaviour and, in my opinion, should be Fx's default behavior as well.

As mentioned, this is the default behavior of pg_dump and essential to load a schema that includes functions. Without this, creating a functions that depend on tables that are yet to be created fails (ie: will be created in a subsequent query from the schema).

teoljungberg:
Thanks @cmer - I need to do some reading on this and testing. Can't promise when or if that happens, but I will try.

cmer:
Thanks! This would be game-changer TBH. Hope you see the value in it.

lib/fx/configuration.rb
```diff
@@ -40,9 +40,25 @@ class Configuration
...
+    # Check function bodies during creation by issuing
+    # `SET LOCAL check_function_bodies TO <<value>>;` before creating functions.
+    #
+    # This is only useful with the Postgresql adapter.
+    #
+    # See https://stackoverflow.com/a/36983831 and
+    # https://www.postgresql.org/docs/9.5/runtime-config-client.html#GUC-CHECK-FUNCTION-BODIES
+    #
+    # Set to `nil` to use the database default configuration.
+    # The default and recommended value is `false` to mimic pg_dump's behavior.
```
amitsaxena, line 52: I am not sure about this. The [docs](https://www.postgresql.org/docs/9.5/runtime-config-client.html#GUC-CHECK-FUNCTION-BODIES) say:
> This parameter is normally on.

So maybe the default should be `true` here, with the possibility to set it to `false` for specific use cases?

cmer: You could do that. It's off by default for `pg_dump` from what I remember, and this gem is the closest thing to `pg_dump` so it seemed reasonable to follow the `pg_dump` conventions. Either way works. Just having the option would be a huge win!

cmer: See: https://dba.stackexchange.com/questions/284260/how-to-prevent-pg-dump-from-generating-set-default-table-access-method-heap -- it shows that `false` the default value for `pg_dump`.

teoljungberg:
Keeping this open as I'd like to support this, if it proves of value!

I have just released a new version after lagging behind a lot, with some small refactorings and versioning updates (ruby x rails).

What I'd like to tackle next is #180, because it dawned on me that I'm supporting EOL postgres versions and that feels...redundant.

Once that is done, I'll come back to this. I need to review some more what this functional use-case is, benefits, drawbacks. But expect that after the holiday break.

cmer:
I'm still here for it.

PR #140, opened 2023-12-14
Schema-dump all functions within search_path, not just 'public'

smudge:
For an application relying on `schema.rb`, this gem will dump only functions in the hardcoded `public` schema. This means that an application with tables in other pg schemas (and/or an application with a nonstandard schema search path) may be able to create a function with `create_function` that does not survive the round-trip to `schema.rb`, in turn breaking the app after `db:schema:load`.

One workaround is to use `structure.sql` (which relies on `pg_dump`) to dump the functions in all schemas. However, because the postgres `search_path` is accessible from within a query, this PR proposes using that value to dump all functions within the current `search_path`. This allows the end-developer to continue using `schema.rb` while inserting functions into multiple PG schemas, and is backwards compatible with the prior `= 'public'` check (assuming `public` is in the search path, which it is by default).

**Alternatives considered:**
- Filter functions not by namespace, but by owner, and only dump functions owned by the current user.
- Look up grants and only dump (non-information-schema) functions granted to the current user (or to the PUBLIC role).
- Use `pg_dump` (or Rails' code that executes `pg_dump`) and extract functions. However, this implementation actually uses `search_path` to construct the `--schema=` argument for the command, which is how I came full circle back to using `search_path` in this PR. It actually matches what Rails does in its Ruby logic for producing `structure.sql`:
    ```ruby
    args += search_path.split(",").map do |part|
      "--schema=#{part.strip}"
    end
    ```
    ([source](https://github.com/rails/rails/blob/66e5ed7fa94fcc972729707c7218fa061b715cff/activerecord/lib/active_record/tasks/postgresql_database_tasks.rb#L63-L64))

### DIFF
--- a/lib/fx/adapters/postgres/functions.rb
+++ b/lib/fx/adapters/postgres/functions.rb
@@ -19,7 +19,17 @@ module Fx
               ON pd.objid = pp.oid AND pd.deptype = 'e'
           LEFT JOIN pg_aggregate pa
               ON pa.aggfnoid = pp.oid
-          WHERE pn.nspname = 'public' AND pd.objid IS NULL
+          WHERE pn.nspname
+              IN (
+                  SELECT replace(
+                      unnest(string_to_array(setting, ', ')),
+                      '"$user"',
+                      current_user
+                  )
+                  FROM pg_settings
+                  WHERE name = 'search_path'
+              )
+              AND pd.objid IS NULL
               AND pa.aggfnoid IS NULL
           ORDER BY pp.oid;
         EOS
--- a/spec/fx/adapters/postgres/functions_spec.rb
+++ b/spec/fx/adapters/postgres/functions_spec.rb
@@ -29,5 +29,93 @@ RSpec.describe Fx::Adapters::Postgres::Functions, :db do
         $function$
       EOS
     end
+
+    context "when functions are in a different schema" do
+      it "does not return the `Function` objects" do
+        connection = ActiveRecord::Base.connection
+        connection.execute <<-EOS.strip_heredoc
+          CREATE SCHEMA test_schema;
+          CREATE OR REPLACE FUNCTION test_schema.test()
+          RETURNS text AS $$
+          BEGIN
+              RETURN 'test';
+          END;
+          $$ LANGUAGE plpgsql;
+        EOS
+
+        functions = Fx::Adapters::Postgres::Functions.new(connection).all
+
+        expect(functions).to be_empty
+      end
+
+      context "when the other schema is in the search path" do
+        it "returns the `Function` objects" do
+          connection = ActiveRecord::Base.connection
+          connection.execute <<-EOS.strip_heredoc
+            CREATE SCHEMA test_schema;
+            SET search_path TO public,test_schema;
+            CREATE OR REPLACE FUNCTION test_schema.test()
+            RETURNS text AS $$
+            BEGIN
+                RETURN 'test';
+            END;
+            $$ LANGUAGE plpgsql;
+          EOS
+
+          functions = Fx::Adapters::Postgres::Functions.new(connection).all
+
+          first = functions.first
+          expect(functions.size).to eq 1
+          expect(first.name).to eq "test"
+          expect(first.definition).to eq <<-EOS.strip_heredoc
+            CREATE OR REPLACE FUNCTION test_schema.test()
+             RETURNS text
+             LANGUAGE plpgsql
+            AS $function$
+            BEGIN
+                RETURN 'test';
+            END;
+            $function$
+          EOS
+        ensure
+          ActiveRecord::Base.connection.execute 'SET search_path TO "$user", public'
+        end
+      end
+
+      context 'when the other schema is the "$user" (dynamic) schema' do
+        it "returns the `Function` objects" do
+          connection = ActiveRecord::Base.connection
+          current_user = connection.execute("SELECT current_user").first["current_user"]
+          connection.execute <<-EOS.strip_heredoc
+            CREATE SCHEMA #{current_user};
+            SET search_path TO "$user",public;
+            CREATE OR REPLACE FUNCTION #{current_user}.test()
+            RETURNS text AS $$
+            BEGIN
+                RETURN 'test';
+            END;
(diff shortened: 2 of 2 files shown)


lib/fx/adapters/postgres/functions.rb
```diff
@@ -19,7 +19,17 @@ class Functions
               ON pd.objid = pp.oid AND pd.deptype = 'e'
           LEFT JOIN pg_aggregate pa
               ON pa.aggfnoid = pp.oid
-          WHERE pn.nspname = 'public' AND pd.objid IS NULL
+          WHERE pn.nspname
+              IN (
+                  SELECT replace(
+                      unnest(string_to_array(setting, ', ')),
+                      '"$user"',
```
smudge, line 26: I'm not seeing that Rails goes as far as resolving `"$user"` to whatever the current postgres username is when constructing its `--schema=` arguments for `structure.sql`. So this PR maybe goes a little beyond Rails in that respect, and I could buy unwinding this particular case and not supporting functions resolved by this special dynamic schema flag.

teoljungberg: I wonder if we can rewrite this SQL, and if this is time time to do it in a different way. This SQL was written pretty haphazardly by myself without knowing much how I would access these functions. I got it to work and moved on...

Now - I scoured through Scenic which served as a lot of inspiration for fx, and specifically how they resolve their views and filter them out:

https://github.com/scenic-views/scenic/blob/main/lib/scenic/adapters/postgres/views.rb

They don't access anything relative to the owner of the function, and I wonder if we can do something similar here.

Could you look into that and see what we can find to improve this?

They also differentiate defining views as `namespace.my_view` when namespace is no the default `public` - that could also be a thing fx could adapt.

teoljungberg (commented):
Sorry for the late review, holidays and vacations came between.

Thank you for contributing!

lib/fx/adapters/postgres/functions.rb
```diff
@@ -39,7 +42,12 @@ def initialize(connection)
         #
         # @return [Array<Fx::Function>]
         def all
-          functions_from_postgres.map { |function| to_fx_function(function) }
+          functions_from_postgres.map do |function|
+            to_fx_function(
+              "name" => schema_aware_name(function),
+              "definition" => schema_aware_definition(function)
```
smudge, line 48: Thanks for the nudge @teoljungberg! I'm also a `scenic` user and so I dug into how that gem works, and ended up proposing this change on their end (which so far seems to be well-received, though I cannot say whether it will ultimately merge): https://github.com/scenic-views/scenic/pull/401

In resolving some discrepancies within that gem, I ended up relying on the way that a core Rails contribution enabled PostgreSQL-handling for enums: https://github.com/rails/rails/pull/45740

I think it would make a ton of sense for _both_ `fx` and `scenic` to behave similar to this core `enum` approach, which is roughly as follows:
1. Query for all user-created structures that are accessible to the current PG user (regardless of search_path)
2. If the structure's schema does not match `current_schema()`, include the fully-qualified name in `schema.rb`
3. If the structure's schema _does_ match `current_schema()`, do not include the schema in the generated `schema.rb` name.

Points 2 and 3 there are really crucial. This allows a `schema.rb` to be shared across contexts where the "default" search path might change (and is entirely dependent on external/config factors -- it is not guaranteed to be `public`, and some PAAS providers do not even place this in the end-developer's control), while still allowing developers to explicitly specify non-default schemas and have those survive the round-trip to `schema.rb`.

lib/fx/adapters/postgres/functions.rb
```diff
@@ -53,6 +61,25 @@ def functions_from_postgres
         def to_fx_function(result)
           Fx::Function.new(result)
         end
+
+        def schema_aware_name(function)
```
smudge, line 65: I considered moving both of these "schema-aware" mappers to `Fx::Function`, but it also complicates the tests for that class in ways that didn't feel necessary. (I figured I would leave it up to you @teoljungberg, as to whether that class should be a "dumb" wrapper or a "smart" translator.)

lib/fx/adapters/postgres/functions.rb
```diff
@@ -53,6 +61,25 @@ def functions_from_postgres
...
+          if function.fetch("schema") == function.fetch("current_schema")
+            function.fetch("name")
+          else
+            "#{function.fetch("schema")}.#{function.fetch("name")}"
+          end
+        end
+
+        def schema_aware_definition(function)
```
smudge, line 73: The reason this method is necessary is because PG itself is not schema-aware when returning the function definition, and will, for example, include the fully-qualified name (`public.something`) even if `current_schema()` matches the function's schema.

This is probably the correct behavior on PG's part, but it means that for `schema.rb` purposes, we should strip it. (See [here](https://github.com/teoljungberg/fx/pull/140/files#r1459966705) for the reasoning -- this how Rails handles enum names in a way that enables a single `schema.rb` to be valid & shareable in many contexts that may not be configured to have the same default PG schema.)

lib/fx/adapters/postgres/functions.rb
```diff
@@ -10,16 +10,19 @@ class Functions
...
+            current_schema() AS current_schema
           FROM pg_proc pp
           JOIN pg_namespace pn
               ON pn.oid = pp.pronamespace
           LEFT JOIN pg_depend pd
               ON pd.objid = pp.oid AND pd.deptype = 'e'
           LEFT JOIN pg_aggregate pa
               ON pa.aggfnoid = pp.oid
-          WHERE pn.nspname = 'public' AND pd.objid IS NULL
+          WHERE pn.nspname NOT IN ('pg_catalog', 'information_schema')
```
smudge, line 24: Some strategies go as far as ignoring everything in `pg_*`. I'm considering making that change still.

(Unfortunately I have not found any other way to isolate user-created functions, and there are many examples that take a similar approach to what I have done here.)

joelind: @smudge I think you can use [this](https://github.com/teoljungberg/fx/compare/master...joelind:fx:dump-functions-from-all-schemas) (I stupidly branched and created a PR before I saw this one)

PR #146, opened 2024-05-25
Move connection-specific code

teoljungberg:


teoljungberg:
To follow up - why does this fail locally and not on CI?

PR #156, opened 2024-11-07
Add optional alphabetical ordering in schema dump

salidux:
Active Record’s schema dumping convention is to output all elements in a consistent, name-sorted order. However, F(x) currently fetches functions and triggers by their insertion order (oid), which can lead to order inconsistencies in db/schema.rb when working with database dumps from production or staging.

Restoring from a pg_dump, which outputs functions and triggers alphabetically, often causes unnecessary schema churn when regenerating db/schema.rb locally:

- Engineer 1 adds function `a()` alongside an existing function `b()`, resulting in `db/schema.rb` listing them as `b(), a()`.
- Engineer 2, needing a production data refresh, restores their database from a pg_dump, which reorders the functions as `a(), b()`.
- Engineer 1 continues working locally, regenerating `db/schema.rb` to `b(), a()` again.

The new `dump_functions_and_triggers_alphabetically` configuration provides an option to mitigate this. When enabled, functions and triggers are ordered alphabetically at the `SchemaDumper`. This ensures compatibility across database adapters while maintaining the default behavior if the option is not set.

By matching the alphabetical order pattern of `db/schema.rb`, this configuration reduces schema churn for teams working with database dumps across different environments.

I know this was already discussed in https://github.com/teoljungberg/fx/pull/139, but this approach is a bit different. First, it's an optional parameter, and secondly, this approach would work with any DB adapter.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
       fail-fast: false
       matrix:
         ruby: ["3.0", "3.1", "3.2", "3.3", "3.4.0-preview2"]
-        rails: ["6.1", "7.0", "7.1", "7.2", "8.0.0.beta1"]
+        rails: ["6.1", "7.0", "7.1", "7.2", "8.0.0.rc1"]
         continue-on-error: [false]
 
     services:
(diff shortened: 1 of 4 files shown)


teoljungberg:
I don't expect the sorting to handle when a developer uses a database dump from a staging environment.

While I agree that it is not ideal to sort by insertion order, I don't see a reason to change it due to using database dumps.

As previously mentioed - I'd love to support topological sorting. And have the sorting be consistent that way, and allow functions and triggers to depend on others.

PR #159, opened 2024-11-18
Feature/added support for apartment gem

abdullahmajid-devsinc:
* By default the gem add function with public as schema in schema.rb. At the time of creating a new tenant, system tries to load schema but fails to find method.

* Added a tweak to support Apartment Gem by appending public. to function in trigger in schema.rb

---------


teoljungberg:
As I'm not using Apartment for any project, I'm not keen on supporting this.
Please publish this as a separate gem instead, or patch it inside your application.

However - I'd entertain PRs that allows for customization and modification to support such a feature.

PR #160, opened 2024-11-26
Add support for multi db setups with a different adapter

### LINKED ISSUE #155: Support for multiple databases
I'm using Rails 8 multiple databases. For primary I'm using postgres, but for cache and queues I'm using SQLite.

When trying to run migrations for _those_ databases, this gem looks to be intercepting

```
$ rails db:migrate:cache --trace
bin/rails aborted!
ActiveRecord::StatementInvalid: SQLite3::SQLException: no such table: pg_proc (ActiveRecord::StatementInvalid)
Caused by:
SQLite3::SQLException: no such table: pg_proc (SQLite3::SQLException)
Tasks: TOP => db:schema:dump:errors
```

For now I've had to remove this gem as I can't figure out what monkey patch I need to do to try working around it. That said, I think it's similar to things mentioned here:
https://github.com/scenic-views/scenic/issues/402

manuelvanrijn:
I was implementing solid_cable using SQLite as an adapter and came across this issue, where the schema dumper was trying to dump the functions and triggers for the SQLite solid cable schema. Obviously, this isn't possible.

I've modified the code to hook into the `ActiveRecord::ConnectionAdapters::PostgreSQL::SchemaDumper` instead of the generic `ActiveRecord::SchemaDumper` and added a test

This should also fix #155

### DIFF
--- a/fx.gemspec
+++ b/fx.gemspec
@@ -27,5 +27,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "activerecord", ">= 7.0"
   spec.add_dependency "railties", ">= 7.0"
 
+  spec.add_development_dependency "sqlite3"
+
   spec.required_ruby_version = ">= 3.0"
 end
--- a/lib/fx.rb
+++ b/lib/fx.rb
@@ -21,7 +21,7 @@ module Fx
   def self.load
     ActiveRecord::Migration::CommandRecorder.include(Fx::CommandRecorder)
     ActiveRecord::ConnectionAdapters::AbstractAdapter.include(Fx::Statements)
-    ActiveRecord::SchemaDumper.prepend(Fx::SchemaDumper)
+    ActiveRecord::ConnectionAdapters::PostgreSQL::SchemaDumper.prepend(Fx::SchemaDumper)
 
     true
   end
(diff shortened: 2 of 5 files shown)


teoljungberg:
Fx is postgresql first, but there are adapters for the other databases that will break as a result of this.

Which I’m not interested doing a deprecation cycle for.

I still agree with the issue, but this solution wouldn’t cut it.

I don’t have any other ideas in mind, as I don’t have a rails application running on any other database than postgresql

manuelvanrijn:
For now, I'm using the following to be able to use both Postgres as SQLite:

```ruby
module Fx
  module Adapters
    class Null
      def initialize(*)
      end

      def functions = []
      def triggers = []

      %i[create_function create_trigger update_function update_trigger drop_function drop_trigger
         execute connection].each do |method_name|
        define_method(method_name) { |*| nil }
      end
    end
  end

  def self.database
    connectable = ActiveRecord::Base
    adapter_class = if connectable.connection.adapter_name == "PostgreSQL"
                      Fx::Adapters::Postgres
                    else
                      Fx::Adapters::Null
                    end
    adapter_class.new(connectable)
  end
end
```

/cc @MaksJS

If the naming of all the adapter classes were equal to the name's `ActiveRecord::ConnectionAdapters` are using (for instance `PostgreSQLAdapter`), we could have implemented it in fx.

Thanks

PR #165, opened 2025-02-11
Adds Rails 6.1 if_exists flags

humphreyja:
Rails 6.1 added `if_exists` flags to things like `remove_column` to assist with edge cases of migrations where the column was removed and possibly not readded in a rollback; basically putting your entire migration in a state of failure. Just ran across this with this gem, so this seems like a good addition.

Blog post showing how the `if_exists` flag works.
https://blog.saeloun.com/2020/02/10/rails-support-for-if-exists-if-not-exists-on-remove-column-add-column-in-migrations

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,11 +77,17 @@ tweak the schema in the new definition and run the `update_function` migration.
 
 ## I don't need this trigger or function anymore. Make it go away.
 
-F(x) gives you `drop_trigger` and `drop_function` too:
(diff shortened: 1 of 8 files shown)


teoljungberg:
The standardrb warnings is something I'd like to see get dealt with in this PR. I'll continue with refactoring once this is merged as I have some stylistic ideas that aren't on you as the contributor to deal with.

PR #168, opened 2025-08-11
Scope functions and triggers to schema search path

### LINKED ISSUE #164: Support functions in different schemas / search_paths
Hello, it would be lovely if this gem supported different schemas. We use a different schema than `public`, by setting `search_path` in various places when we run SQL; so our existing functions don't exist in `public` (and therefore don't get dumped). 

I am monkey patching by updating `public` on [this line](https://github.com/teoljungberg/fx/blob/master/lib/fx/adapters/postgres/functions.rb#L22):
```
WHERE pn.nspname = 'public' AND pd.objid IS NULL
```
to the main schema we use.

tylerhunt:
Updates the function and trigger queries for PostgreSQL to only return objects from the current schema search path.

This is based on how Rails limits dumping objects like enums (rails/rails#47791).

Resolves #164.

### DIFF
--- a/lib/fx/adapters/postgres/functions.rb
+++ b/lib/fx/adapters/postgres/functions.rb
@@ -19,7 +19,8 @@ module Fx
               ON pd.objid = pp.oid AND pd.deptype = 'e'
           LEFT JOIN pg_aggregate pa
               ON pa.aggfnoid = pp.oid
-          WHERE pn.nspname = 'public' AND pd.objid IS NULL
+          WHERE pn.nspname = ANY (current_schemas(false))
+              AND pd.objid IS NULL
               AND pa.aggfnoid IS NULL
           ORDER BY pp.oid;
         EOS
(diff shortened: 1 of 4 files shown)


teoljungberg:
Hm, the tests are indeed flaky after this. See #169 as I'm trying to reproduce it without pushing more commits to master.

Any ideas as to how to fix this @tylerhunt ?

tylerhunt:
The specs seem to have been flaky to begin with. Running `rake` passes fine, but the order in which the specs run appears to be important, as RSpec's order randomization causes issues (e.g. `rspec --seed 40439`). These issues appear on `master` even before this PR was merged (tested with 9edd1da7d6cafcc345efe8c50572aef81f0da479). It's probably not a trivial task to make them idempotent, and I don't know how far back the issue goes.

teoljungberg:
Huh, I have not run into this. I wonder what caused it.

frank-west-iii:
> ### Pull request successfully merged and closed
> You're all set — the branch has been merged.

Hey @teoljungberg, any plan to release a new version soon? We are adding fx and having to monkey patch for this same reason. Thanks!

PR #169, opened 2025-08-11
WIP - Debug CI flakyness

teoljungberg:

PR #170, opened 2025-08-11
Improve spec resiliency to resolve random failures

tylerhunt:
This should help eliminate some of the spec failures that can occur due to differences in the order in which the specs are executed.


tylerhunt:
Updated to account for existing `my_table`.

teoljungberg:
I'm continuing to debug in https://github.com/teoljungberg/fx/pull/169, and ensuring that the schema path is set to its default at the beginning of each test "seems enough".

I'll rerun it a few times and see, 3 runs green so far.

teoljungberg:
Now 5 successful runs, I'll consider that a success.

Pushed as edb9da0

Closing this in favor of that patch, but thank you @tylerhunt for contributing and debugging!

tylerhunt:
I'm still seeing random failures after multiple runs locally. I think this PR should be merged because it does address actual issues with the spec suite.

teoljungberg:
I can't reproduce it on CI (trying with https://github.com/teoljungberg/fx/pull/171 with a set of extra reruns) nor locally.

spec/fx/adapters/postgres/triggers_spec.rb
```diff
@@ -38,13 +38,12 @@
       expect(first.definition).to include("EXECUTE FUNCTION uppercase_users_name()")
 
       connection.execute "CREATE SCHEMA IF NOT EXISTS other;"
-      connection.execute "SET search_path = 'other';"
 
-      triggers = Fx::Adapters::Postgres::Triggers.new(connection).all
-
-      expect(triggers).to be_empty
+      connection.with schema_search_path: "other" do
```
tylerhunt, line 42: This change (and the corresponding one in `functions_spec.rb`) should also do away with the need to reset the search path `around :each` (edb9da0).

teoljungberg: Yeah the whole thing is strange to me, but the resetting at the beginning of each test is pretty OK for me as a solution.

PR #177, opened 2025-10-28
ensure multi-schema dumping in Rails 8.1.0 doesn't dump the same objects for all schemas

agrobbin:
In Rails 8.1.0, `bin/rails db:schema:dump` now knows how to dump multiple schemas (rails/rails#50020). To ensure we don't dump the same functions / triggers for each schema, we need to stop memoizing `Fx.database.{functions,triggers}`.

This should be fully backwards compatible to previous Rails versions, as we still only call each of those methods once per schema dumping invocation, it just now works if schema dumping is invoked multiple times.

### DIFF
--- a/lib/fx/schema_dumper.rb
+++ b/lib/fx/schema_dumper.rb
@@ -4,14 +4,12 @@ module Fx
     def tables(stream)
       if Fx.configuration.dump_functions_at_beginning_of_schema
         functions(stream)
-        empty_line(stream)
       end
 
       super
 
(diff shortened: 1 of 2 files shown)


teoljungberg:
Can you add a test to document this behavior under rails 8.1?

teoljungberg:
Can we confirm that this fails on Rails 8.1 without the patch? Just for sake of coverage?

teoljungberg:
Excellent, thank you for confirming!

PR #195, opened 2026-01-11
Add spaces between functions and triggers in schema.rb

chiperific:
I've been dealing with the changes to schema_dumper in Rails 8.1 and Senic 1.9 and found that it's most common for their methods not to emit a newline after their last operation.

Hence I think it's safe to assume that functions() and triggers() should emit a newline before their loops and within their loops, up to the last iteration.
And, not having these functions emit a final newline holds with the pattern used by the Rails and Scenic schema_dumpers.

I've tested this in my own apps with a patch on this gem and the spacing is correct with the new Rails 8.1 schema changes (and with Scenic emitting `create_view` statements before the `create_function` and `create_trigger` statements).

### DIFF
--- a/lib/fx/schema_dumper.rb
+++ b/lib/fx/schema_dumper.rb
@@ -20,7 +20,7 @@ module Fx
     def functions(stream)
       dumpable_functions_in_database = Fx.database.functions
 
-      dumpable_functions_in_database.each_with_index do |function, index|
+      dumpable_functions_in_database.each do |function|
         stream.puts
         stream.puts(function.to_schema)
       end
@@ -29,7 +29,7 @@ module Fx
     def triggers(stream)
       dumpable_triggers_in_database = Fx.database.triggers
 
-      dumpable_triggers_in_database.each_with_index do |trigger, index|
+      dumpable_triggers_in_database.each do |trigger|
         stream.puts
         stream.puts(trigger.to_schema)
       end


teoljungberg:
I don't quote understand this patch, what it aims to solve, without a test or similar showcasing why.

lib/fx/schema_dumper.rb
```diff
@@ -20,22 +20,26 @@ def tables(stream)
...
       dumpable_functions_in_database = Fx.database.functions
 
-      dumpable_functions_in_database.each do |function|
+      return unless dumpable_functions_in_database.any?
+
+      stream.puts
+
+      dumpable_functions_in_database.each_with_index do |function, index|
         stream.puts(function.to_schema)
+        stream.puts unless index == dumpable_functions_in_database.size - 1
```
teoljungberg: I'd flip this around and do

```
  dumpable_functions_in_database.each_with_index do |function, index|
    stream.puts if index > 0
    stream.puts(function.to_schema)
  end
```

Or even use `.positive?`

lib/fx/schema_dumper.rb
```diff
@@ -20,22 +20,26 @@ def tables(stream)
     def functions(stream)
       dumpable_functions_in_database = Fx.database.functions
 
-      dumpable_functions_in_database.each do |function|
+      return unless dumpable_functions_in_database.any?
```
teoljungberg: I'd swap this

return if dumpable_functions_in_database.blank?`
```suggestion
      return if dumpable_functions_in_database.blank?
      
```

lib/fx/schema_dumper.rb
```diff
@@ -20,22 +20,26 @@ def tables(stream)
...
-        stream.puts
-      end
+      return unless dumpable_triggers_in_database.any?
+
+      stream.puts
 
-      dumpable_triggers_in_database.each do |trigger|
+      dumpable_triggers_in_database.each_with_index do |trigger, index|
         stream.puts(trigger.to_schema)
+        stream.puts unless index == dumpable_triggers_in_database.size - 1
```
teoljungberg: Same, as the prior comment.

chiperific:
> I don't quote understand this patch, what it aims to solve, without a test or similar showcasing why.

Thanks for the review. I'll add a test to show where spaces are being emitted.

chiperific:
@teoljungberg, made those changes, ready for another review.

teoljungberg:
I added an extra test to document when there is nothing to dump as b0b9c3f

Issue #197, opened 2026-02-11
Sort functions by dependency


mhenrixon:
I have this in an initializer to make sorting correctly. Without it, things are created in alphabetical order and not in the order they are required:

```ruby
# frozen_string_literal: true

if defined?(Fx)
  Fx.configure do |config|
    # Ensures that functions are dumped at the beginning of the schema file.
    config.dump_functions_at_beginning_of_schema = true
  end

  # Topologically sort functions so dependencies are created first.
  # The default fx query orders by OID (creation order), which can place
  # a function that references another function before its dependency.
  # e.g. `euclidian` references `vec_sub` but was created first.
  module Fx
    module DependencySortedSchemaDumper
      private

      def functions(stream)
        all_functions = Fx.database.functions
        sorted = topological_sort_functions(all_functions)

        sorted.each do |function|
          stream.puts(function.to_schema)
        end
        stream.puts if sorted.any?
      end

      def topological_sort_functions(functions)
        names = functions.to_set(&:name)
        # Build adjacency: function A depends on function B if A's definition
        # mentions B by name (as a word boundary match).
        deps = {}
        functions.each do |fn|
          deps[fn.name] = functions.select { |other|
            other.name != fn.name &&
              fn.definition.match?(/\b#{Regexp.escape(other.name)}\s*\(/)
          }.map(&:name)
        end

        sorted = []
        visited = {}

        visit = lambda { |fn_name|
          return if visited[fn_name] == :done
          return if visited[fn_name] == :in_progress # cycle guard

          visited[fn_name] = :in_progress
          (deps[fn_name] || []).each do |dep|
            visit.call(dep)
          end
          visited[fn_name] = :done
          sorted << fn_name
        }

        functions.each do |fn|
          visit.call(fn.name)
        end

        # Rebuild in sorted order, preserving duplicates (overloaded functions)
        name_order = sorted.each_with_index.to_h
        functions.sort_by { |fn| [name_order[fn.name] || 0, fn.definition] }
      end
    end

    # Sort triggers by name for deterministic output.
    # The default fx query orders by table OID, which varies between databases.
    module DeterministicTriggerSchemaDumper
      private

      def triggers(stream)
        dumpable_triggers = Fx.database.triggers
        return unless dumpable_triggers.any?

        stream.puts
        dumpable_triggers.sort_by(&:name).each do |trigger|
          stream.puts(trigger.to_schema)
        end
      end
    end

[...]

PR #200, opened 2026-02-22
Add configurable function sorter for schema dumper

teoljungberg:
- Add config.function_sorter to allow custom function ordering
  in schema.rb
- Any callable responding to .call(functions) can be assigned
  to control the order functions are written during schema dump
- Defaults to nil (preserves existing database order)

---

/cc @mhenrixon -- this provides the hook you need from #197.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ changelog, see the [commits] for each version via the version links.
   schema.rb (#197)
 - Internal refactorings / improvements
   - Add PostgreSQL 18 to CI test matrix
+  - Simplify conditional function dumping in SchemaDumper (#200)
 
 ## [0.10.2]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ changelog, see the [commits] for each version via the version links.
 - Add PostgreSQL versioning policy, officially supporting PostgreSQL 14-18
 - Refactor Statements module to use explicit keyword arguments instead of `**options` hash (#186)
 - Add `config.function_sorter` to allow custom function ordering in
-  schema.rb (#197)
+  schema.rb (#197, #200)
 - Internal refactorings / improvements
   - Add PostgreSQL 18 to CI test matrix
   - Simplify conditional function dumping in SchemaDumper (#200)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,6 @@ changelog, see the [commits] for each version via the version links.
   schema.rb (#197, #200)
 - Internal refactorings / improvements
   - Add PostgreSQL 18 to CI test matrix
-  - Simplify conditional function dumping in SchemaDumper (#200)
 
 ## [0.10.2]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ changelog, see the [commits] for each version via the version links.
 
 - Add PostgreSQL versioning policy, officially supporting PostgreSQL 14-18
 - Refactor Statements module to use explicit keyword arguments instead of `**options` hash (#186)
+- Add `config.function_sorter` to allow custom function ordering in
+  schema.rb (#197)
 - Internal refactorings / improvements
   - Add PostgreSQL 18 to CI test matrix
 
--- a/lib/fx/configuration.rb
+++ b/lib/fx/configuration.rb
@@ -16,7 +16,7 @@ module Fx
     attr_accessor :dump_functions_at_beginning_of_schema
 
     # A callable that sorts functions before they are dumped to schema.rb.
-    # Must respond to `#call(functions)` and return a sorted array of
+    # Must respond to `.call(functions)` and return a sorted array of
     # {Fx::Function} objects.
     #
     # Defaults to nil (no sorting, preserves database order).
--- a/lib/fx/configuration.rb
+++ b/lib/fx/configuration.rb
@@ -15,9 +15,18 @@ module Fx
     # @return [Boolean] Boolean
     attr_accessor :dump_functions_at_beginning_of_schema
 
+    # A callable that sorts functions before they are dumped to schema.rb.
+    # Must respond to `#call(functions)` and return a sorted array of
+    # {Fx::Function} objects.
+    #
+    # Defaults to nil (no sorting, preserves database order).
+    # @return [#call, nil] Function sorter
+    attr_accessor :function_sorter
+
     def initialize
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -11,8 +11,11 @@ changelog, see the [commits] for each version via the version links.
 
 - Add PostgreSQL versioning policy, officially supporting PostgreSQL 14-18
 - Refactor Statements module to use explicit keyword arguments instead of `**options` hash (#186)
+- Add `config.function_sorter` to allow custom function ordering in
+  schema.rb (#197, #200)
 - Internal refactorings / improvements
   - Add PostgreSQL 18 to CI test matrix
+  - Simplify conditional function dumping in SchemaDumper (#200)
```
teoljungberg:
```suggestion
```

oskarpearson:
I'd love to see this released (my monkey-patch that sorted functions and triggers broke on the most recent release).

Would it be possible to extend this to enable sorting of triggers too?

Could I suggest adding an example to the config file or README? (the README just refers to the configuration file, but an example in the README like this might be useful? Then it the config info be removed from https://github.com/teoljungberg/fx/blob/928dfa9822f5776f60ee561312f5f53976ffd8e7/lib/fx.rb#L47

```
# Configuration

If you want to change from the default configuration, add a `config/initializers/fx.rb` file as follows:

Fx.configure do |config|
  # Set the database adapter. Currently only Fx::Adapters::Postgres (the default) is supported
  # config.database = Fx::Adapters::Postgres

  # Default is to save functions at the end of the schema. Set to 'true' to move them to the top
  # config.dump_functions_at_beginning_of_schema = false

  # Default is save functions and triggers to schema.rb in the underlying database order.
  # If you want to sort them instead, provide a callable that takes an array of Fx::Function / Fx::Trigger
  # objects and then returns them in the required order
  # config.function_sorter = ->(functions) { functions.sort_by(&:name) }
  # config.trigger_sorter = ->(functions) { functions.sort_by(&:name) }
end
```

Thank you!

teoljungberg:
> I'd love to see this released (my monkey-patch that sorted functions and triggers broke on the most recent release).

Oh interesting! What broke so I can be more careful next time?

> Would it be possible to extend this to enable sorting of triggers too?

Sure thing, I meant to give flexibility into how things are sorted / placed. Unsure I will ever cover "how" to do each of them in the fear of messing things up for the many when providing something for the few

> Could I suggest adding an example to the config file or README? (the README just refers to the configuration file, but an example in the README like this might be useful? Then it the config info be removed from 

Of course, when / if this is released it will be documented

teoljungberg:
If you need further patching-entrypoints I'd patch https://github.com/teoljungberg/fx/blob/928dfa9822f5776f60ee561312f5f53976ffd8e7/lib/fx/adapters/postgres/functions.rb#L33 with siblings as those are public API.

teoljungberg:
This and https://github.com/teoljungberg/fx/pull/210 makes me think if this is worthwhile doing. The public API of the adapters should (and will) be kept to what they are. I wonder if this is enough to instead be able to...

```ruby
class SortedPostgresAdapter < Fx::Adapters::Postgres
  def functions
    super.sort_by(&:name)
  end

  def triggers
    super.sort_by(&:name)
  end
end

Fx.configure do |config|
  config.database = SortedPostgresAdapter.new
end
```

I've yet to decide.

oskarpearson:
> > I'd love to see this released (my monkey-patch that sorted functions and triggers broke on the most recent release).
> 
> Oh interesting! What broke so I can be more careful next time?

Oh, don't be more careful about private interfaces :) It should be all at my risk, after all, but thank you!

a77ebcdb5e25bbbbd91f7d4decde0b953a56da40 removed the `dumpable_functions_in_database` and `dumpable_triggers_in_database` methods, and I'd monkey patched them as follows:

```
Fx::SchemaDumper.class_eval do
  private

  def dumpable_functions_in_database
    @dumpable_functions_in_database ||= Fx.database.functions.sort_by(&:name)
  end

  def dumpable_triggers_in_database
    @dumpable_triggers_in_database ||= Fx.database.triggers.sort_by(&:name)
  end
end
```

oskarpearson:
> This and #210 makes me think if this is worthwhile doing. The public API of the adapters should (and will) be kept to what they are. I wonder if this is enough to instead be able to...
> 
> ```ruby
> class SortedPostgresAdapter < Fx::Adapters::Postgres
>   def functions
>     super.sort_by(&:name)
>   end
> 
>   def triggers
>     super.sort_by(&:name)
>   end
> end
> 
> Fx.configure do |config|
>   config.database = SortedPostgresAdapter.new
> end
> ```

This method works perfectly, by the way. I've removed my monkeypatch.

Thanks!

teoljungberg:
> > > I'd love to see this released (my monkey-patch that sorted functions and triggers broke on the most recent release).
> > 
> > 
> > Oh interesting! What broke so I can be more careful next time?
> 
> Oh, don't be more careful about private interfaces :) It should be all at my risk, after all, but thank you!
> 
> [a77ebcd](https://github.com/teoljungberg/fx/commit/a77ebcdb5e25bbbbd91f7d4decde0b953a56da40) removed the `dumpable_functions_in_database` and `dumpable_triggers_in_database` methods, and I'd monkey patched them as follows:
> 
> ```
> Fx::SchemaDumper.class_eval do
>   private
> 
>   def dumpable_functions_in_database
>     @dumpable_functions_in_database ||= Fx.database.functions.sort_by(&:name)
>   end
> 
>   def dumpable_triggers_in_database
>     @dumpable_triggers_in_database ||= Fx.database.triggers.sort_by(&:name)
>   end
> end
> ```

Gotcha, those were indeed private.

Happy to hear the adapter patch is working for you!

mhenrixon:
Still haven't gotten around to test it out. 

I was kind of hoping to be able to test the default dependency sort with it.

teoljungberg:
@mhenrixon I'm uncertain what I want to bundle with the gem as this is uncharted territory for me. That is why I'm erring on the side of providing configuration, or means by extension, to test this out in real life applications before rolling it out.

I would appreciate if you, and others, can use the approach from ~~#210~~ #211 and then the spiked out dependency sorter in #206 in your applications and see how that works!

teoljungberg:
Moving this to draft in favor of #211, and waiting testing of #211 + #206 before proceeding

teoljungberg:
Closing in favor of the adapter subclass pattern documented in #211. The adapter approach is more flexible and avoids new API surface. Will revisit if there's a strong need for a dedicated config option.

PR #203, opened 2026-03-01
Auto-detect function argument types when dropping or updating

### LINKED ISSUE #7: Can't use function with parameters

teoljungberg:
Functions with parameters could not be dropped or updated
because PostgreSQL requires argument types in DROP FUNCTION.
This has been the longest-standing open issue (#7, 2017).

- Look up argument types from pg_proc automatically at
  drop/update time
- Accept an explicit arguments: option on drop_function and
  update_function for overloaded functions
- Raise Fx::AmbiguousFunctionError with guidance when multiple
  overloads exist and no arguments: option is given
- Forward arguments: through Fx::Statements using
  **options.slice(:arguments), keeping custom adapters working

Closes #7

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ changelog, see the [commits] for each version via the version links.
   functions, fixing support for functions with parameters (#7)
 - Add `arguments:` option to `drop_function` and `update_function` for
   targeting specific overloads of functions that share a name
+- Custom adapters that override `drop_function` or `update_function` must
+  now accept an `arguments:` keyword argument
 - Internal refactorings / improvements
   - Add scheduled EOL check for Ruby, Rails, and PostgreSQL (#205)
   - Add GitHub release creation to release task (#209)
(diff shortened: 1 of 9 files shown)


carter-thaxton:
This looks great.  I like that it handles the unambiguous case automatically, and provides a straightforward way to handle overloads.

What will it take to get this into the released version of the gem?  I'd love to start using it right away.

PR #206, opened 2026-03-07
Add built-in function sorters for dependency ordering

teoljungberg:
Add two built-in function sorters for config.function_sorter
(from #200) to order functions by dependency in schema.rb:

- FunctionsSortByDependency: parses SQL definitions with regex
  to detect function calls. Works without a DB connection.
  Handles PL/pgSQL well since it scans function bodies
  directly. Strips comments and string literals to avoid false
  positives.
- FunctionsSortByPgDepend: queries PostgreSQL's pg_depend
  catalog for real dependency edges. Authoritative for
  SQL-language functions. Does not cover PL/pgSQL (PostgreSQL
  defers parsing to execution time).

Both use TSort with strongly_connected_components to tolerate
mutually recursive functions.


teoljungberg:
Closing in favor of the adapter subclass pattern documented in #211. The adapter approach is more flexible and avoids new API surface. Will revisit if there's a strong need for built-in sorters.

PR #210, opened 2026-03-13
Add config.trigger_sorter for schema dumper

teoljungberg:
- Add `config.trigger_sorter` to allow custom trigger ordering in schema.rb, mirroring the existing `config.function_sorter`


teoljungberg:
Closing in favor of the adapter subclass pattern documented in #211. The adapter approach is more flexible and avoids new API surface. Will revisit if there's a strong need for a dedicated config option.

PR #213, opened 2026-05-04
Add worked TSort example for dependency-ordered function dumps

jeffdlange:
## Summary

The "Customizing Schema Dump Order" section added in #211 mentions that Ruby's `TSort` can be used for dependency-based ordering, but doesn't show how.

This PR adds a complete, copy/pasteable, example to the README.

## What it does

Subclasses `Fx::Adapters::Postgres` and overrides `#functions` to:

1. Discover function-to-function dependencies by scanning `pg_proc.prosrc` textually for whole-word references to other managed function names.
2. Run the resulting graph through `TSort#tsort` so functions are emitted after every function they reference.

## Why textual scanning, not `pg_depend`

`pg_depend` looks like the natural source of truth here, but PostgreSQL only records function-to-function dependencies for new-style (PG14+) SQL functions written with `RETURN ...` or `BEGIN ATOMIC ... END;`, whose body is stored as a parsed query tree.

The `CREATE OR REPLACE FUNCTION ... LANGUAGE sql AS $$ ... $$` form that F(x) generates (and `LANGUAGE plpgsql` functions in general) store the body as opaque text in `pg_proc.prosrc` with no `pg_depend` rows for inter-function references. So a `pg_depend`-based topological sort would be a no-op for F(x)-generated functions.

The textual `prosrc` scan works uniformly for `LANGUAGE sql`, `LANGUAGE plpgsql`, and any other language whose body is stored in `prosrc`.

## Why this matters

The default order returned by the database (`ORDER BY pp.oid`, i.e. creation order) is not guaranteed to be dependency-safe. After out-of-order migrations across branches, or after a function is dropped and recreated, the on-disk OID order can flip — at which point `schema.rb` load fails with `function X(...) does not exist` because `LANGUAGE sql` resolves identifiers at `CREATE` time.

## Notes

- The example uses the documented adapter-subclass extension point. No new gem API is added.
- Filters in the dependency query mirror the ones F(x) already uses for its own dump query (`current_schemas(false)`, no extension-owned functions, no aggregates), so the graph is built over the same set of functions F(x) is about to dump.
- Built-in/extension function names that appear in a body (`sum`, `round`, `unnest`, …) produce harmless edges that get dropped when the graph looks them up and finds no matching node.
- The heredoc uses single-quoted marker (`<<~'SQL'`) so Ruby doesn't interpret the PostgreSQL word-boundary anchors `\m` / `\M` as escape sequences.

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,6 +136,110 @@ The same approach works for more advanced ordering. For example, if your
 functions depend on each other and need to be dumped in dependency order, you
 could use Ruby's built-in `TSort` to topologically sort them.
 
+### Topological sort for dependent functions
+
+If your functions reference each other, dump-order matters: a function written
+in `LANGUAGE sql` is name-resolved at `CREATE` time, so a referenced function
+must already exist when its caller is created. The default order returned by
+the database is not guaranteed to be dependency-safe — for example, after
+out-of-order migrations across branches, or after dropping and recreating a
+function, the on-disk creation order may have flipped.
+
+`pg_depend` looks like the natural source of truth here, but PostgreSQL only
+records function-to-function dependencies for new-style (PG14+) SQL functions
+written with `RETURN ...` or `BEGIN ATOMIC ... END;`, whose body is stored as
+a parsed query tree. The `CREATE OR REPLACE FUNCTION ... LANGUAGE sql AS $$
+... $$` form that F(x) generates (and `LANGUAGE plpgsql` functions in general)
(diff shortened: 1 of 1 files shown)


teoljungberg:
Hello and thanks for the PR! There is prior work someone can search for through the issues (#197) and the attempts I did (#206 and 201).

I don't think we need to highlight any of them until the need for this is high enough, and then we'll start with shipping it as a separate gem, then making it an opt in behavior somehow.

jeffdlange:
Understood, thanks for taking the time!

